PR #1, opened 2022-11-14
WAL-266

BtencateSphereon:


App.tsx
```diff
@@ -13,8 +13,16 @@ import store from './src/store'
...
-  'Require cycle:', // Ignore require cycles for the app in dev mode. They do show up in Metro!
-  'Non-serializable values were found in the navigation state' // https://reactnavigation.org/docs/troubleshooting/#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state
+  // Ignore require cycles for the app in dev mode. They do show up in Metro!
+  'Require cycle:',
+  /*
+    Non-serializable values were found in the navigation state. Check:
+    This can break usage such as persisting and restoring state. This might happen if you passed non-serializable values such as function, class instances etc. in params. If you need to use components with callbacks in your options, you
+    can use 'navigation.setOptions' instead. See https://reactnavigation.org/docs/troubleshooting#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state for more details.
+   */
+  'Non-serializable values were found in the navigation state',
```
nklomp, line 23: Hmzz, not sure I want to ignore these errors. They only end up in debug builds and we really should do something about these (they are in our control)

src/components/labels/SSIStatusLabel/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <Container style={[style, { borderColor: color }]}>
+      {showIcon && <BadgeContainer>{getStatusBadge(status, color)}</BadgeContainer>}
+      <StatusCaption style={{ color }}>{getStatusTranslation(status)}</StatusCaption>
+    </Container>
+  )
+}
+
+const getStatusTranslation = (status: LabelStatus): string => {
+  switch (status) {
```
nklomp, line 34: Not sure about this, but let's keep it for now. We can always refactor these into separate methods

nklomp (changes requested):
Overall good. As long as Sergey has seen a demo. I do believe we shouldn't ignore the added LogBox items. They are in our control and only anny us developers in develop builds ;)
The cycle thing is more out of our control

src/components/labels/SSIStatusLabel/index.tsx
BtencateSphereon: I agree it is not the best option, but i wanted to merge all the status components we had as they were copies. And i need to stop refactoring stuff else this ticket will not get done. And i already did what we always said we should not do and that is pull in a lot of refactor work which is outside of the scope of the ticket...sorry :)

App.tsx
BtencateSphereon: I do not believe this is an issue for us. As we do not use the state in the way they describe. I added a todo to move the actions we pass in to navigation.setOptions

nklomp: I do believe we use this anti-pattern though (passing complex object around):
https://reactnavigation.org/docs/params#what-should-be-in-params

nklomp: What is your opinion on that?

nklomp: Also the 2nd ignored error either needs a ticket or it needs to be removed from this PR.

BtencateSphereon: It is a fine line. I think we already discussed to set a credentialId in the redux state to indicate which credential is being used in flows. That would solve one of these issues. For the actions, i do like passing them into screens as we have screens with multiple purposes and it solves issues we had in other mobile apps where screens got really messy.

Besides that, we use a lot of functional components and you do not want to tie those down. So for initial screens sure. but once we get to functional components other rules will count

BtencateSphereon: Added a todo for second error

nklomp:
LGTM

PR #2, opened 2022-11-16
WAL-335

BtencateSphereon:
needs a rebase


hrehman-sphereon (approved):
We could add a spinner. But basic objective of the ticket is achieved.

nklomp:
A generic loading animation needs to be created anyway.  But that is out of scope here

PR #4, opened 2022-11-23
feat: Match actual keys on kids from the DID. Use latest openid4vci client

nklomp:
This PR is about:
- Using the new OpenID4VCI client class for the Open ID Provider
- Do actual matching of authentication VM relationships on the proper key in Veramo. So the kid in the PoP reflects the proper key, instead of making assumptions about it being the first key of the DID document
- Cleanups and fixes


BtencateSphereon (approved):
LGTM

PR #15, opened 2022-12-13
WAL-366

BtencateSphereon:
PR using local artefacts, need a new release of the ssi-sdk


sksadjad (approved):
looks good to me, although we still need to change the local packages

PR #17, opened 2022-12-27
WAL-294

sanderPostma:


package.json
```diff
@@ -31,7 +31,6 @@
     "@sphereon/ssi-sdk-lto-did-provider": "0.8.1-unstable.86",
     "@sphereon/ssi-sdk-vc-api-issuer": "0.8.1-unstable.86",
     "@sphereon/ssi-types": "0.8.1-unstable.86",
-    "@unimodules/core": "^7.1.2",
```
nklomp, line 34: AFAIK, we explicitly add this for OpenID4VCI support, which is vital in the app. Have you tested the app to be working with one of the issuers from the README?

patches/@bitauth+libauth+1.19.1.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/node_modules/@bitauth/libauth/build/main/lib/format/log.js b/node_modules/@bitauth/libauth/build/main/lib/format/log.js
+index 6a1f908..0fedcfc 100644
+--- a/node_modules/@bitauth/libauth/build/main/lib/format/log.js
```
nklomp, line 3: Since this patch is disabling functionality instead of patching it, please provide at least a comment

nklomp (changes requested):
See remarks

package.json
BtencateSphereon: Seems this package is deprecated

[npm: @unimodules/core](https://www.npmjs.com/package/@unimodules/core)

And has been replaced by the expo package

https://blog.expo.dev/whats-new-in-expo-modules-infrastructure-7a7cdda81ebc

So i think we need to look into replacing it.

android/app/build.gradle
```diff
@@ -247,7 +247,7 @@ android {
         release {
             // Caution! In production, you need to generate your own keystore file.
             // see https://reactnative.dev/docs/signed-apk-android.
-            signingConfig signingConfigs.debug
+            // signingConfig signingConfigs.debug
```
BtencateSphereon, line 250: what is the reason for commenting this out? would it make sense to remove it if we do not need it?

android/sign.cmd
```diff
@@ -0,0 +1 @@
+jarsigner -keystore .\eburger-play-key\upload-keystore.jks .\app\build\outputs\bundle\release\app-release.aab eburger-upload-keystore
```
BtencateSphereon, line 1: this is referencing eburger :)

ios/Podfile
```diff
@@ -35,21 +35,21 @@ target 'SSIWallet' do
 #    pod 'Permission-BluetoothPeripheral', :path => "#{permissions_path}/BluetoothPeripheral"
```
BtencateSphereon, line 35: would it make sense to just remove the ones we do not need?

BtencateSphereon (dismissed):
left a few comments

package.json
sanderPostma: The ios compiler errors out because duplicate symbols when unimodules is in the dependencies, because it's already present by default in the expo package. Not sure if OpenID4VCI will still be able to import it, or that is has stopped working now.

patches/@bitauth+libauth+1.19.1.patch
sanderPostma: Not sure whether it is possible to write comments in .patch files. A detailed explanation is in WAL-294.

android/app/build.gradle
sanderPostma: The singer started to complain about it. We don't need a debug signature in a release packages. Removed.

android/sign.cmd
sanderPostma: Yes we are using the same Play store account and the same key store / upload key. Back in the day someone named them to be eBurger, but should be just "Sphereon". Renaming...

sanderPostma: The keystore itself is not in git. The alias in it is still "eburger-upload-keystore" though. There might be a way to rename that, but this seems kind of trivial.

ios/Podfile
sanderPostma: Yes, but at the time I had no idea which ones we are actually using in the app, still don't actually. I just commented out as many as possible so Apple would stop whining. Waiting for feedback whether the app is fully functional, maybe some have to be enabled still.

patches/@bitauth+libauth+1.19.1.patch
nklomp: Thr point is that we need something to link it from the actual code. Your patch is literally commenting some code out, of course you then can add a comment in that commented out code, explaining why or linking to that ticket

sanderPostma: Done " --> this regex expression does not work in RN, see WAL-294 for details"

package.json
nklomp: Okay let's see how this plays out

nklomp (approved):

package.json
BtencateSphereon: There is still an issue met the patches

nklomp:
This PR is now merged and adjusted with the latest updates

PR #21, opened 2023-01-12
Upgrade Expo/React-Native versions to latest

sanderPostma:
DO NOT MERGE until WAL-389 and then WAL-294 are merged into develop


nklomp (changes requested):
Block review because other PRs need to be merged first. I have not reviewed the actual code

PR #22, opened 2023-01-17
WAL-380

BtencateSphereon:


BtencateSphereon:
needs an update for local artefacts

src/services/qrService.ts
```diff
@@ -277,17 +276,16 @@ const connectOpenId4VcIssuance = async (args: IQrDataArgs) => {
       })
 
   const provider = await OpenId4VcIssuanceProvider.initiationFromUri({ uri: args.qrData.uri })
-  provider.getServerMetadataAndPerformCryptoMatching().then((metadata: IServerMetadataAndCryptoMatchingResponse) => {
+  provider.getServerMetadataAndPerformCryptoMatching().then(async (metadata: IServerMetadataAndCryptoMatchingResponse) => {
     const url = new URL(metadata.serverMetadata.issuer)
-    // TODO WAL-380 should be handled by the connection-manager
-    if (!store.getState().contact.contacts.some((contact: IConnectionParty) => contact.name === url.host)) {
-      // TODO fix this type issue
+    const contacts = await getContacts({ filter: [{ name: url.host }] })
```
nklomp, line 281: This should become uri as key with a value of url.protocol and hostname instead of name right? I know the user cannot change the name right now. That is what we have the alias for. But since this is the contact we are setting the uri to the generic hostname already.

The problem is that the user could update the uri I guess. Maybe we should add a baseUri or initialURI or something

nklomp (commented):
Small remark about the search functionality

src/services/qrService.ts
BtencateSphereon: uri is optional as well, so i dont tthink that will work

src/services/connectionService.ts
```diff
@@ -129,18 +140,28 @@ const addDefaultConnections = async () => {
   }
 
   const firm24Name = 'Firm24'
-  const firm24 = parties.find((party: IConnectionParty) => party.name === firm24Name)
-  if (!firm24) {
+  const firm24CorrelationId = 'sphereon.com'
```
nklomp, line 143: It has the same correlationId as sphereon (website is firm24.com)

src/services/qrService.ts
```diff
@@ -277,17 +279,21 @@ const connectOpenId4VcIssuance = async (args: IQrDataArgs) => {
...
 
   const provider = await OpenId4VcIssuanceProvider.initiationFromUri({ uri: args.qrData.uri })
-  provider.getServerMetadataAndPerformCryptoMatching().then((metadata: IServerMetadataAndCryptoMatchingResponse) => {
+  provider.getServerMetadataAndPerformCryptoMatching().then(async (metadata: IServerMetadataAndCryptoMatchingResponse) => {
     const url = new URL(metadata.serverMetadata.issuer)
-    // TODO WAL-380 should be handled by the connection-manager
-    if (!store.getState().contact.contacts.some((contact: IConnectionParty) => contact.name === url.host)) {
-      // TODO fix this type issue
+    const contacts = await getContacts({ filter: [{ identifier: { correlationId: url.hostname } }] })
+    console.log(`contacts: ${contacts.length}`)
```
nklomp, line 285: console.log

nklomp (changes requested):
Some small remarks. Also need to update the deps to become published deps from other PR

src/services/connectionService.ts
BtencateSphereon: fixed

src/services/qrService.ts
BtencateSphereon: removed

nklomp (approved):
LGTM

PR #23, opened 2023-01-24
WAL-365

BtencateSphereon:


patches/@unimodules+core+7.1.2.patch
```diff
@@ -1,319 +0,0 @@
-diff --git a/node_modules/@unimodules/core/.eslintrc.js b/node_modules/@unimodules/core/.eslintrc.js
```
hrehman-sphereon, line 1: For my info:
Why is this removed?
What was it doing?
Why isn't it required anymore?

.eslintrc.json
```diff
@@ -12,7 +12,7 @@
     "plugin:import/typescript",
     "prettier"
   ],
-  "globals": { "BigInt": true, "console": true, "WebAssembly": true },
+  "globals": { "BigInt": true, "console": true, "WebAssembly": true, "__DEV__": true, "localStorage": true },
```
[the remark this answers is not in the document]
BtencateSphereon: i'm not sure bout that, as this solves an issue about eslint complaining about the shim. What is your reason this should be false?

patches/@unimodules+core+7.1.2.patch
BtencateSphereon: because the package has been removed

.eslintrc.json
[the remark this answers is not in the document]
hrehman-sphereon: Got it. I missunderstood.

yarn.lock
```diff
@@ -3834,6 +3834,21 @@
     "@typescript-eslint/types" "5.45.1"
     eslint-visitor-keys "^3.3.0"
 
+"@unimodules/core@*":
```
hrehman-sphereon, line 3837: As per comment referred by link ... Do we need to remove it from here then ? 

https://github.com/Sphereon-Opensource/ssi-mobile-wallet/pull/23#discussion_r1087732851

patches/@unimodules+core+7.1.2.patch
nklomp: I am not too sure about this. Indeed it is still showing up in the yarn.lock file I see nothing that removes the package. It is being listed as an optional dep from isomorphic-webcrypto. 

So the question is. Where has this been explicitly removed and what is preventing its installation then?

BtencateSphereon: @nklomp this happened in sanders pull request

hrehman-sphereon (approved):

nklomp (approved):

PR #31, opened 2023-02-13
[WAL-421] Remove default connections

hrehman-sphereon:


src/services/connectionService.ts
```diff
@@ -142,56 +142,4 @@ const addDefaultConnections = async () => {
       )
     })
   }
-
-  const firm24Name = 'Firm24'
```
BtencateSphereon, line 146: this only removes the firm24 connection. We want to remove both connections and the call to addDefaultConnections. just remove the whole function

BtencateSphereon (changes requested):

src/services/connectionService.ts
hrehman-sphereon: Removed the whole function now.

BtencateSphereon (approved):

PR #32, opened 2023-02-13
[WAL-419] show correct credential status in credential details screen.

hrehman-sphereon:


src/utils/DateUtils.ts
```diff
@@ -24,6 +25,13 @@ class DateUtils {
       DateUtils.DATE_FORMAT_OPTIONS
     )
   }
+
+  static getCredentialStatus(expirationDate: number): CredentialStatusEnum {
```
BtencateSphereon, line 29: would move this to a src/utils/CredentialUtils

src/utils/mappers/CredentialMapper.ts
```diff
@@ -1,6 +1,7 @@
 import { ICredential } from '@sphereon/ssi-types'
 
-import { CredentialStatusEnum, ICredentialDetailsRow, ICredentialSummary } from '../../@types'
+import { ICredentialDetailsRow, ICredentialSummary } from '../../@types'
+import DateUtils from "../DateUtils";
```
BtencateSphereon, line 4: run yarn fix:prettier

src/utils/DateUtils.ts
```diff
@@ -24,6 +25,13 @@ class DateUtils {
...
+    return !expirationDate || expirationDate >= new Date().valueOf() / 1000
```
BtencateSphereon, line 30: just an improvement we could make, instead of the 1000 value that was already there. we should use DateUtils.EPOCH_MILLISECONDS

src/utils/DateUtils.ts
hrehman-sphereon: Moved to CredentialUtils.
No need to make the interface we are determining the status of the credential. The credential is now used as param.

src/utils/mappers/CredentialMapper.ts
hrehman-sphereon: fixed

src/utils/DateUtils.ts
hrehman-sphereon: fixed.

src/utils/CredentialUtils.ts
```diff
@@ -0,0 +1,44 @@
+import {ICredential} from "@sphereon/ssi-types";
+import {ICredentialStatus} from "@sphereon/ssi-types/src/types/vc";
+
+import {CredentialStatusEnum, ICredentialSummary} from "../@types";
+
+class CredentialUtils {
```
nklomp: Please just create functions instead of creating classes with only static methods. This is not Java ;)

hrehman-sphereon: Got it. Fixed.

src/utils/mappers/CredentialMapper.ts
```diff
@@ -56,10 +57,8 @@ export function toCredentialSummary(verifiableCredential: ICredential, hash?: st
     ? new Date(verifiableCredential.expirationDate).valueOf() / 1000
     : 0
   const issueDate = new Date(verifiableCredential.issuanceDate).valueOf() / 1000
-  const credentialStatus =
-    !expirationDate || expirationDate >= new Date().valueOf() / 1000
-      ? CredentialStatusEnum.VALID
-      : CredentialStatusEnum.EXPIRED
+
+  const credentialStatus = getCredentialStatus(verifiableCredential)
   //TODO add revoked status support
```
BtencateSphereon, line 62: this comment can be moved to to isRevoked function, or can be removed now

src/utils/CredentialUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+    // This should be done in a separate ticket.
+    // {
+    //  id: 'https://revocation-sphereon.sphereon.io/services/credentials/wallet-dev#2022021400',
+    //  type: 'RevocationList2022021401Status',
+    //  revocationListIndex: '2022021402',
+    //  revocationListCredential: 'https://revocation-sphereon.sphereon.io/services/credentials/wallet-dev#2022021400',
+    // }
+  }
+
+export const isExpired = (expirationDate: string | number | undefined) : CredentialStatusEnum | undefined => {
```
BtencateSphereon, line 30: i would expect this method to return if a credential is expired (boolean) and then use that value to return a certain status. Then we can reuse this function a lot more.

Same for the isRevoked

src/utils/mappers/CredentialMapper.ts
hrehman-sphereon: removed.

BtencateSphereon (approved):

PR #39, opened 2023-02-21
feature/WAL-429/del_button

hrehman-sphereon:


nklomp (changes requested):
Isn't this simply removing all delete functionality including that of VCs?

Also the ticket has a TODO, to link it to a new ticket to implement contact deletion. You are now referring to a ticket that is already done, which isn;'t the desired outcome of the ticket

hrehman-sphereon:
@nklomp https://sphereon.atlassian.net/browse/WAL-410 May be some incorrect link was checked. This ticket seem to be in TODO. And this is the ticket mentioned in the Jira Ticket.

I fixed the delete credential as well.

src/components/views/SSISwipeRowViewItem/index.tsx
```diff
@@ -12,15 +12,16 @@ export interface IProps {
   viewItem: JSX.Element
   listIndex: number
   onPress: () => Promise<void>
-  onDelete: () => Promise<void>
+  onDelete?: () => Promise<void>
```
BtencateSphereon, line 15: add a comment here to revert this optional behavior, WAL-410 as this onDelete should be mandatory

hrehman-sphereon: Added.

BtencateSphereon (approved):

nklomp (approved):

PR #40, opened 2023-02-21
feature/WAL-430/hide_activity_tab

hrehman-sphereon:


src/screens/SSIContactDetailsScreen/index.tsx
```diff
@@ -18,7 +18,7 @@ enum ContactTabRoutesEnum {
 }
 
 const SSIContactDetailsScreen: FC<Props> = (props: Props): JSX.Element => {
-  const { contact } = props.route.params
```
BtencateSphereon, line 21: i'm not a fan of this solution. we do not need this option here and i now need to turn it on. like discussed. all you need to do is comment out the routes in this screen. that way they do not get rendered.

src/screens/SSIContactDetailsScreen/index.tsx
```diff
@@ -2,7 +2,7 @@ import { NativeStackScreenProps } from '@react-navigation/native-stack'
 import React, { FC } from 'react'
 
 import SSIActivityView from '../../components/views/SSIActivityView'
```
BtencateSphereon, line 4: if this one is not used, it should be commented out as well

BtencateSphereon (approved):

PR #42, opened 2023-02-22
WAL-422

BtencateSphereon:


src/@config/credentials/contexts/kyc-v1.jsonld
```diff
@@ -0,0 +1,8 @@
+{
```
nklomp, line 1: This one can be deleted

src/@config/credentials/contexts/profile-v1.jsonld
```diff
@@ -0,0 +1,8 @@
+{
```
nklomp, line 1: Can be deleted

src/@config/credentials/contexts/socialmedia-v1.jsonld
```diff
@@ -0,0 +1,7 @@
+{
```
nklomp, line 1: can be deleted

src/agent/index.ts
```diff
@@ -61,17 +70,18 @@ export const didProviders = {
 }
 
 const dbConnection = getDbConnection(DB_CONNECTION_NAME)
+const privateKeyStore: PrivateKeyStore = new PrivateKeyStore(dbConnection, new SecretBox(DB_ENCRYPTION_KEY))
 
 const agent = createAgent<
-  IDIDManager & IKeyManager & IDataStore & IDataStoreORM & IResolver & IDidAuthSiopOpAuthenticator & IConnectionManager
+  IDIDManager & IKeyManager & IDataStore & IDataStoreORM & IResolver & IDidAuthSiopOpAuthenticator & IConnectionManager & ICredentialHandlerLDLocal
```
nklomp, line 76: I am missing ICredentialssuer (probably you are using our VC hanlder ld local directly to issue credentials

src/agent/index.ts
```diff
@@ -118,4 +138,5 @@ export const sendSiopAuthorizationResponse = agent.sendSiopAuthorizationResponse
 export const keyManagerSign = agent.keyManagerSign
 export const dataStoreGetVerifiableCredential = agent.dataStoreGetVerifiableCredential
 export const dataStoreDeleteVerifiableCredential = agent.dataStoreDeleteVerifiableCredential
+export const createVerifiableCredentialLDLocal = agent.createVerifiableCredentialLDLocal
```
nklomp, line 141: I really dislike these aliases and think we should stop using them. It also has problems with autocomplete and your IDE not picking up when methods are renamed etc.

nklomp, line 141: We really should use the Veramo methods and not directly call the LD local method. In the future we also want to be able to issue JWT credentials, which ICredentialIssuer/Veramo already supports. The only difference is that you will have to add a param that denotes the VC type (lds for JsonLD and jwt for JWT)

src/services/credentialService.ts
```diff
@@ -23,3 +30,7 @@ export const getVerifiableCredential = async (args: IGetVerifiableCredentialArgs
 export const deleteVerifiableCredential = async (args: IDeleteVerifiableCredentialArgs): Promise<boolean> => {
   return dataStoreDeleteVerifiableCredential({ hash: args.hash })
 }
+
+export const createVerifiableCredential = async (args: ICreateVerifiableCredentialArgs): Promise<VerifiableCredentialSP> => {
```
nklomp, line 34: We really should use the Veramo methods and not directly call the LD local method. In the future we also want to be able to issue JWT credentials, which ICredentialIssuer/Veramo already supports. The only difference is that you will have to add a param that denotes the VC type (lds for JsonLD and jwt for JWT)

src/store/actions/onboarding.actions.ts
```diff
@@ -44,6 +47,32 @@ export const finalizeOnboarding = (): ThunkAction<Promise<void>, RootState, unkn
             identifier: contactIdentifier
           })
         )
+        dispatch(
+          createVerifiableCredential({
+            credential: {
+              "@context": [
+                "https://www.w3.org/2018/credentials/v1",
+                "https://www.w3.org/2018/credentials/examples/v1",
```
nklomp, line 55: Don't include the examples in here. Only the credentials/v1 and our context should be in here

src/store/actions/onboarding.actions.ts
```diff
@@ -44,6 +47,32 @@ export const finalizeOnboarding = (): ThunkAction<Promise<void>, RootState, unkn
...
+              id: uuidv4(),
+              type: [
+                "VerifiableCredential",
+                "SphereonWalletIdentity"
+              ],
+              issuer: identifier.did,
+              issuanceDate: new Date(),
+              credentialSubject: {
+                id: identifier.did,
+                SphereonWalletIdentity: {
```
nklomp, line 67: I would put the names/email directly in credential subject. No need for an aditional object

nklomp (changes requested):
See remarks,

Main issue is not using Veramo's credential issuer (which delegates to VC handler LD local). See the example code I posted in slack. We need that, as we also want to support JWT credentials for issuance and verification

nklomp (approved):

nklomp (approved):
LGTM

PR #43, opened 2023-02-24
WAL-379

BtencateSphereon:


src/services/storageService.ts
```diff
@@ -27,29 +27,84 @@ export const getPin = async (): Promise<string> => {
...
+  })
+  .catch((error: Error) =>
+    Promise.reject(new Error(`Unable to set value for key: ${STORAGE_PIN_KEY}. ${error.message}`))
+  )
 
-  const users: Map<string, IUser> = JSON.parse(users_value)
-  users.set(user.id, user)
 
-  return AsyncStorage.setItem(STORAGE_USERS_KEY, JSON.stringify(users))
+  // const result: string | null = await AsyncStorage.getItem(STORAGE_USERS_KEY)
```
sksadjad: if we don't need these code block, it's better to delete them, and if we do need them at a later point, please make a comment

sksadjad (changes requested):
looks good, I just added a small request for a commented block of code

src/services/storageService.ts
BtencateSphereon: i think i made the commit when you wrote your comment :P

sksadjad (approved):
LHTM

PR #44, opened 2023-02-24
[WAL-432] Profile Icon

hrehman-sphereon:


src/components/assets/icons/SSIEntityIcon/index.tsx
```diff
@@ -1,29 +1,40 @@
...
+import { connect } from 'react-redux'
 
 import { entities, fonts } from '../../../../styles/colors'
+import { RootState } from '../../../../types'
+import { getGeneralInitialsOfUserName } from '../../../../utils/UserUtils'
 
 export interface IProps {
   size?: number
-  color?: ColorValue
+  fontColour?: ColorValue
```
BtencateSphereon, line 12: Color instead of colour

src/components/assets/icons/SSIEntityIcon/index.tsx
```diff
@@ -1,29 +1,40 @@
+import { Avatar } from '@rneui/themed'
```
BtencateSphereon, line 1: i rather not use third party packages for this. making a view with a style that makes it a circle does not require anything

BtencateSphereon (changes requested):

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: fixed

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: fixed

src/components/assets/icons/SSIEntityIcon/index.tsx
```diff
@@ -1,29 +1,41 @@
...
+  SSITextH2LightStyled as AvatarText
+} from '../../../../styles/components'
+import { RootState } from '../../../../types'
+import { getDefaultUserNameInitials } from '../../../../utils/UserUtils'
 
 export interface IProps {
   size?: number
-  color?: ColorValue
-  backgroundColor?: ColorValue
+  fontColor?: ColorValue
```
hrehman-sphereon, line 15: I think we should leave it as prop and not move it to styles as in future it is possible to have multiple avatars displayed in same view/list etc.

src/components/assets/icons/SSIEntityIcon/index.tsx
```diff
@@ -1,29 +1,41 @@
...
+  bgColor?: ColorValue
```
hrehman-sphereon, line 16: I think we should leave it as prop and not move it to styles as in future it is possible to have multiple avatars displayed in same view/list etc.

src/components/assets/icons/SSIEntityIcon/index.tsx
```diff
@@ -1,29 +1,41 @@
...
 }
 
-export default SSIEntityIcon
+const mapStateToProps = (state: RootState) => {
+  return {
+    initialsOfUserName: getDefaultUserNameInitials(state.user.activeUser?.firstName, state.user.activeUser?.lastName)
+  }
+}
+
+export default connect(mapStateToProps, null)(SSIEntityIcon)
```
hrehman-sphereon, line 41: I did not fully understand where do we want to move it ? I mean if I pass it as prop from Header then header should also receive it from header's parent. Which will trigger a long chain. I mean we will be passing initials from like 5th - 10th parent of the avatar ... perhaps we can discuss further. As this part is not clear to me.

src/utils/UserUtils.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * This is a helper function to help the wallet user see prefilled initials.
+ *
+ * @param namePart1 is first part of the username. Also known as firstName
+ * @param namePart2 is 2nd part of the username. Also known as lastName, surname or remaining part of the name etc.
+ *
+ * @return the initials selected by default
+ */
+export const getDefaultUserNameInitials = (namePart1?: string, namePart2?: string): string => {
```
hrehman-sphereon, line 11: Named changed.

src/components/assets/icons/SSIEntityIcon/index.tsx
```diff
@@ -1,29 +1,41 @@
 import React, { FC } from 'react'
```
Brummos, line 1: lets rename this component to SSIProfileIcon, to maker it more clear what this component is for

src/components/assets/icons/SSIEntityIcon/index.tsx
Brummos, line 16: backgroundColor

src/components/assets/icons/SSIEntityIcon/index.tsx
Brummos: ye, i guess that using the redux store is justified here to avoid passing it as a route prop everytime.

so lets keep the connect. look at examples where we use connect, that component is an extension of a PureComponent. Use that structure here as well.

src/components/assets/icons/SSIEntityIcon/index.tsx
```diff
@@ -1,29 +1,41 @@
...
-          fill={color}
-        />
-      </Svg>
-    </View>
+    <Avatar
```
Brummos, line 24: for consistancy, name this SSIProfileIconBackground

src/components/assets/icons/SSIEntityIcon/index.tsx
```diff
@@ -1,29 +1,41 @@
...
+      style={{
+        backgroundColor: bgColor,
+        width: size
+      }}
+    >
+      <AvatarText style={{ color: fontColor }}>{initialsOfUserName}</AvatarText>
```
Brummos, line 30: for consistancy, name this SSIProfileIconText

src/styles/components/components/SSIEntityIcon/index.ts
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components/native'
+
+export const SSIEntityIconStyled = styled.View`
+ margin-top: 5px;
```
Brummos, line 4: i would not expect this margin to be present as this will mover the circle 5 pixels downwards. any reason why we do this?

src/utils/UserUtils.tsx
Brummos: i am not sure what is happening here, i would expect 1 function called getInitials that gets 1 string, which is the full name of the user. and returns 2 letters

src/utils/UserUtils.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * This is a helper function to help the wallet user see prefilled initials.
+ *
+ * @param namePart is part of the username
+ *
+ * @return the partial initials selected by default
+ */
+export const getPartialInitials = (namePart?: string): string => {
```
Brummos, line 29: this feels like a lot of code just to get 2 letters out of a full name

src/utils/UserUtils.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * This is a helper function to help the wallet user see prefilled initials.
+ *
+ * @param namePart is part of the username
+ *
+ * @return the partial initials selected by default i.e. first alphabet last word
+ */
+function getFirstAlphabetOfLastWord(namePart?: string): string {
```
Brummos, line 45: this feels like a lot of code just to get 2 letters out of a full name

src/components/assets/icons/SSIEntityIcon/index.tsx
```diff
@@ -1,29 +1,41 @@
...
+    </Avatar>
   )
 }
 
-export default SSIEntityIcon
+const mapStateToProps = (state: RootState) => {
+  return {
+    initialsOfUserName: getDefaultUserNameInitials(state.user.activeUser?.firstName, state.user.activeUser?.lastName)
```
Brummos, line 37: wrong use of the mapStateToProps here. look at another place where we user this to get the activeUser. then use the data on this object to call the utils function within the renderer not within mapStateToProps 

const mapStateToProps = (state: RootState) => {
  return {
    activeUser: state.user.activeUser
  }
}

then you can use this as props.activeUser within the code

Brummos (changes requested):

src/components/assets/icons/SSIEntityIcon/index.tsx
Brummos: using mapStateToProps has the benefit of when we set an new activeUser, this will automatically update as it triggers a rerender

src/styles/components/components/SSIEntityIcon/index.ts
```diff
@@ -0,0 +1,9 @@
...
+ align-items: center;
+ justify-content: center;
+ aspect-ratio: 1;
+ border-radius: 17px;
```
Brummos, line 8: what happends with this when i set the size of the icon to, lets say 50? will it still be round? My guess is, it wont :). for now i think we can set a fixed size for the profile icon. so move the width to this style component

src/components/assets/icons/SSIEntityIcon/index.tsx
```diff
@@ -1,29 +1,41 @@
...
-  color?: ColorValue
-  backgroundColor?: ColorValue
+  fontColor?: ColorValue
+  bgColor?: ColorValue
+  initialsOfUserName: string
 }
 
 const SSIEntityIcon: FC<IProps> = (props: IProps): JSX.Element => {
-  const { size = 34, color = fonts.light, backgroundColor = entities['100'] } = props
+  const { size = 34, fontColor = fonts.light, bgColor = entities['100'], initialsOfUserName = '?' } = props
```
Brummos, line 21: lets rename initialsOfUserName  to initials, as there are no other initials

src/utils/UserUtils.tsx
Brummos: just a quick search. something like this but maybe a little cleaner:

unction getInitials(fullName) {
  const nameArray = fullName.trim().split(" "); // split the full name into an array of individual names
  const firstInitial = nameArray[0][0]; // get the first letter of the first name
  const lastInitial = nameArray[nameArray.length - 1][0]; // get the first letter of the last name
  return `${firstInitial}${lastInitial}`; // combine the two initials into a string and return it
}

// Example usage:
const fullName = "John Smith Jr.";
const initials = getInitials(fullName);
console.log(initials); // output: "JS"

src/components/assets/icons/SSIEntityIcon/index.tsx
Brummos: these can be kept here as they are dynamically set

src/utils/UserUtils.tsx
hrehman-sphereon: We do not have any rule for what can be in the first and the last name the only checks are that it should not be empty.

So what is happening here is we Just select 2 alphabets ... These function just provide a user default initials. Even when the user enters a 
FirstName: '.@Sam
LastName: __--Pretty

The current implementation will just find SP. That's why so much is happening in the code.

nklomp: That will fail miserably with non alphabet names. Please simply trimm everything.  Split by whitespace. Take the first and last word and uppercase their first letters

hrehman-sphereon: @nklomp that is exactly what I am doing ... And it is definitely working ... It's working for even weired inputs like all special characters. 
It will return a string '?'. But it will always respond with a usable string.

nklomp: So why are the regexes in there then that match the alphabet?

Creating an avatar should not be concerned with any input validation. If a user can enter '#$% 123' the result should be #1 in the avatar

src/components/assets/icons/SSIEntityIcon/index.tsx
Brummos: i see i pasted the wrong thing, omit SSI

src/components/assets/icons/SSIEntityIcon/index.tsx
Brummos: i see i pasted the wrong thing, omit SSI

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: I am confirming that we did not need to convert the functional component to the pure component, right? we just needed to follow the connect pattern used in pure component?

src/utils/UserUtils.tsx
```diff
@@ -0,0 +1,7 @@
+
+export const getInitials = (fullName: string): string => {
+  const nameArray = fullName.trim().split(" ");
+  const firstInitial = nameArray[0][0];
+  const lastInitial = nameArray[nameArray.length - 1][0]; // get the first letter of the last name
```
nklomp, line 5: Should be a bit more defensive. Check at least the arrays after splitting as this is a public function.

Also the values should be uppercased

src/@config/constants/index.ts
```diff
@@ -22,3 +22,5 @@ export const PIN_CODE_LENGTH = 6
 
 export const EMAIL_ADDRESS_VALIDATION_REGEX = /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/
 export const ONLY_ALLOW_NUMBERS_REGEX = /^\d+$/
+export const FIRST_ALPHABET_REGEX = /[a-zA-Z]/
```
BtencateSphereon, line 25: regexes still being used?

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: retained

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: Connect has been used differently now.

src/utils/UserUtils.tsx
hrehman-sphereon: After personally talking to Niels last week the @Brummos suggestion was opted.

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: renamed

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: changed

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: improved

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: improved

src/styles/components/components/SSIEntityIcon/index.ts
hrehman-sphereon: Removed the margin

src/utils/UserUtils.tsx
hrehman-sphereon: Removed the complex code. Made it simple as suggested by @Brummos

src/utils/UserUtils.tsx
hrehman-sphereon: Removed the complex code. Made it simple as suggested by @Brummos

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: fixed.

src/styles/components/components/SSIEntityIcon/index.ts
hrehman-sphereon: moved to the styles for now. The dynamic styling for all sizes has not been added for now.

src/components/assets/icons/SSIEntityIcon/index.tsx
hrehman-sphereon: The variable was removed.

src/utils/UserUtils.tsx
hrehman-sphereon: fixed

src/@config/constants/index.ts
hrehman-sphereon: No, removed.

src/utils/UserUtils.tsx
hrehman-sphereon: @nklomp please mark it resolved.

PR #46, opened 2023-02-27
WAL-340

BtencateSphereon:


src/localization/translations/nl.json
```diff
@@ -134,5 +134,6 @@
     "notifications_overview_empty_state_title": "Geen notificaties",
     "notifications_overview_empty_state_subtitle": "Op dit moment heb je geen notificaties.\nDe notificaties verschijnen hier automatisch.",
     "lock_title": "Log in\nde Sphereon Wallet",
-    "lock_subtitle": "Vul je pincode in."
+    "lock_subtitle": "Vul je pincode in.",
+    "show_raw_credential_button_caption": "Toon rauwe credential"
```
nklomp, line 138: ruwe

nklomp (commented):
I guess we could omit the Show/Toon word from the menu and just have raw/ruwe credential in there. Other than the rauwe->ruwe, LGTM. No need for a re-review after that change

src/localization/translations/nl.json
BtencateSphereon: changed to Ruwe credential

nklomp (approved):

PR #51, opened 2023-03-15
expo fixes

BtencateSphereon:
some fixes for expo,

if the react-native version was meant to be this specific version, then i'll revert that one


BtencateSphereon:
@nklomp, react-native version is reverted

nklomp (approved):

PR #52, opened 2023-03-23
Feature/WAL-409

BtencateSphereon:


nklomp (approved):
Given all the reformatting, it is very hard to do the review. I did look at some of the important classes that have changed and these look okay. Suggest to simply merge it and create tickets/fixes if we encounter any problems

PR #54, opened 2023-03-28
feature/WAL-451

BtencateSphereon:


src/screens/SSICredentialsOverviewScreen/index.tsx
```diff
@@ -72,24 +78,39 @@ class SSICredentialsOverviewScreen extends PureComponent<IProps, IState> {
...
+        title={itemInfo.item.title}
+        issuer={itemInfo.item.issuer}
+        issueDate={itemInfo.item.issueDate}
+        expirationDate={itemInfo.item.expirationDate}
+        credentialStatus={itemInfo.item.credentialStatus}
+        properties={[]}
+      />
+    );
+
+    return this.props.activeUser.identifiers.some((identifier: IUserIdentifier) => identifier.did === itemInfo.item.issuer.name)
```
nklomp, line 94: We should also check for the credential type. Current code basically means that you cannot delete any credential that is issued by the user. In the future we might add other self-asserted credential use cases, not being the wallet identity credential. So filtering on that makes sense

nklomp (changes requested):
We should also check the credential type

src/screens/SSICredentialsOverviewScreen/index.tsx
BtencateSphereon: added a check for the type/title of the vc

nklomp:
LGTM

nklomp (approved):

PR #56, opened 2023-03-30
feature/WAL-479

BtencateSphereon:


src/store/actions/user.actions.ts
```diff
@@ -56,17 +66,31 @@ export const addIdentifier = (args: IAddIdentifierArgs): ThunkAction<Promise<voi
...
-    getUsersFromStorage()
-      .then((users: Map<string, IUser>) => {
+    await getUsersFromStorage()
+      .then(async (users: Map<string, IUser>) => {
         const user = users.get(userId);
         if (user) {
           dispatch({type: SET_ACTIVE_USER_SUCCESS, payload: user});
+          let userState: IUserState = getState().user;
+          while (!userState.activeUser) {
+            await new Promise((resolve) => setTimeout(resolve, 50));
```
nklomp, line 78: Probably wise to add an additional guard here. If it cannot get the user after let's say 10 secs, it will never happen and now we would be stuck in this loop. Adding a guard and then a final check whether the userState is not undefined would be wise.

Similar issue for the contact state below

src/store/actions/user.actions.ts
```diff
@@ -56,17 +66,31 @@ export const addIdentifier = (args: IAddIdentifierArgs): ThunkAction<Promise<voi
...
+            userState = getState().user;
+          }
+          await dispatch(getContacts())
+          let contactState: IContactState = getState().contact;
+          while (contactState.contacts.length === 0) {
```
nklomp, line 83: Please add a remark that the only reason this works is because we always will have at least 1 contact (the user)

nklomp (changes requested):
2 remarks about potential endless loops

src/store/actions/user.actions.ts
BtencateSphereon: added a comment

src/store/actions/user.actions.ts
BtencateSphereon: Added a max timeout for the infinite loop. Even though I am pretty sure the user can never be stuck on this. Setting the active user is not the problem here and that will trigger the navigation to navigate to the home screen. the worst that could happen is that there are no contacts and vc's being displayed. Besides that this action is being dispatched, so it will not stop the app.

nklomp (approved):

nklomp:
LGTM

PR #58, opened 2023-03-31
[WAL-481] add pin code screen before re-opening application

hrehman-sphereon:


src/store/actions/user.actions.ts
```diff
@@ -70,3 +71,10 @@ export const setActiveUser = (userId: string): ThunkAction<Promise<void>, RootSt
       .catch(() => dispatch({type: SET_ACTIVE_USER_FAILED}));
   };
 };
+
+export const removeActiveUser = (): ThunkAction<Promise<void>, RootState, unknown, Action> => {
```
BtencateSphereon, line 75: rename this to logout, and rename the setActiveUser to login. also do this for all the types associated with these actions.

src/store/actions/user.actions.ts
```diff
@@ -70,3 +71,10 @@ export const setActiveUser = (userId: string): ThunkAction<Promise<void>, RootSt
...
+  return async (dispatch: ThunkDispatch<RootState, unknown, Action>) => {
+    dispatch({type: USERS_LOADING});
+    dispatch({type: REMOVE_ACTIVE_USER_SUCCESS, payload: null});
```
BtencateSphereon, line 78: payload should be undefined, not null

src/handlers/LockingHandler/index.ts
```diff
@@ -0,0 +1,32 @@
...
+import {removeActiveUser} from '../../store/actions/user.actions';
+
+class LockingHandler {
+  private lockingEventListener: NativeEventSubscription;
+
+  public enableLocking = async (): Promise<void> => {
+    const handleAppStateChange = async (nextAppState: string): Promise<void> => {
+      if (nextAppState === 'background' || nextAppState === 'active') {
+        debug('Locking application...');
+        console.log('Locking application...');
```
BtencateSphereon, line 19: remove the console log

BtencateSphereon (changes requested):

src/store/actions/user.actions.ts
hrehman-sphereon: @nklomp @Brummos I guess ... making it logout is not right ... Let's say I logged out ... and now want to login with a different user. We have not designed the wallet to handle that... I suggest to keep it the same for now.

In the scope of this 3 story point ticket doing that kind of refactor will increase the testing surface too big ... even the VC and contacts may need to be checked ...

nklomp: It is just a name suggestion, as the removeUser is highly confusing

hrehman-sphereon: renamed

src/store/actions/user.actions.ts
hrehman-sphereon: changed

src/handlers/LockingHandler/index.ts
hrehman-sphereon: removed

BtencateSphereon (approved):
LGTM

PR #71, opened 2023-04-19
feature/WAL-470

BtencateSphereon:


src/components/views/SSICardView/index.tsx
```diff
@@ -77,18 +77,20 @@ const SSICardView: FC<IProps> = (props: IProps): JSX.Element => {
   // The uri is a transparent pixel in case there is not background image
   const backgroundImage = props.backgroundImage
     ? {uri: props.backgroundImage}
-    : {uri: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII='};
+    : {uri: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII='} // TODO which image is this?
```
nklomp, line 80: Just paste that in your browser ;)

This is a single transparent pixel to ensure the error about a null image is gone if no background Image is supplied, which should be clear from the ternary operator IMO

nklomp (approved):
See remark about the background image.

Rest LGTM

src/components/views/SSICardView/index.tsx
BtencateSphereon: ah, ye i did, but could not see anything. :P

PR #73, opened 2023-04-20
[WAL-517] remove add contact screen from stack immediately after cont…

hrehman-sphereon:
…act is created


src/services/qrService.ts
```diff
@@ -248,14 +248,19 @@ const connectOpenId4VcIssuance = async (args: IQrDataArgs): Promise<void> => {
...
+          onCreate: () => onCreateSuccess(metadata),
         });
       } else {
         sendResponseOrSelectCredentials(metadata.credentialsSupported);
       }
     });
   };
 
+  const onCreateSuccess = async (metadata: IServerMetadataAndCryptoMatchingResponse): Promise<void> => {
+    args.navigation.goBack();
```
BtencateSphereon, line 260: technically this is not the solution that was discussed. We discussed to remove the screen from the stack. this navigates back and show the qr scanner again. which is also not the flow maria designed.

look in to calling something like

            args.navigation.reset({
              index: 0,
              routes: [{name: ScreenRoutesEnum.CREDENTIAL_DETAILS}],
            });

when we go to the credential details to reset the stack. this sets the CREDENTIAL_DETAILS as the first screen in the stack so then i would be impossible to naviugate back to the contact add screen.

only issue i ran into is that the navigator lost where it is and the goback is not handled anymore.

BtencateSphereon: args.navigation.reset({
  index: 0,  routes: [{name: ScreenRoutesEnum.QR_READER}, {name: ScreenRoutesEnum.CREDENTIAL_DETAILS}],});

will fix the goback issue

nklomp: Okay, apparently @hrehman-sphereon tried that, but didn't work. I agree that we discussed something else, but since we are already awaiting the contact creation before being able to go back on the next screen, there is no possibility for a race condition, so current approach works and I am fine with that.

nklomp (approved):

src/services/qrService.ts
nklomp: Okay just had a demo and the app is not behaving as you would expect. It quickly shows the QR code reader again, because of the go back in code, which of course is confusing.

src/services/qrService.ts
```diff
@@ -387,6 +394,26 @@ const connectOpenId4VcIssuance = async (args: IQrDataArgs): Promise<void> => {
                 },
               }
             );
+
+            const navigationState = navigationRef.current.getRootState();
+            const { routes, index } = navigationState;
+
+            console.log('   ')
```
nklomp, line 401: Do we still need these console.logs?

nklomp (changes requested):
See remark about the console.logs. For now rest looks okay to me

src/services/qrService.ts
BtencateSphereon: nah, this was some test code to see if i could get the active screens within the stack.

code has been removed

nklomp (approved):
LGTM

PR #75, opened 2023-04-20
altered the code to use new interfaces/methods from openid4vci client & common module

sksadjad:
changes:
- using the new interfaces imported from newly introduced `openid4vci-common` module
- some changes in naming conventions


src/providers/credential/OpenId4VcIssuanceProvider.ts
```diff
@@ -150,7 +151,7 @@ class OpenId4VcIssuanceProvider {
 
   public getCredential = async ({credentialType, pin}: IGetCredentialArgs): Promise<CredentialResponse> => {
     const {issuanceOpts} = await this.getServerMetadataAndPerformCryptoMatching();
-
+    console.log(`in OpenId4VcIssuanceProvider.getCredential issuanceOpts: ${JSON.stringify(issuanceOpts)}`)
```
nklomp, line 154:
```suggestion
```

src/providers/credential/OpenId4VcIssuanceProvider.ts
```diff
@@ -140,6 +140,7 @@ class OpenId4VcIssuanceProvider {
         credentialType,
         pin,
       });
+      console.log(`credentialResponses: ${JSON.stringify(credentialResponses)}`)
```
nklomp, line 143:
```suggestion
```

nklomp (changes requested):
Missing package.json update and also some small console.log lines which can go or moved to debug as it works now

src/providers/credential/OpenId4VcIssuanceProvider.ts
sksadjad: removed that line

src/providers/credential/OpenId4VcIssuanceProvider.ts
sksadjad: removed that line

sksadjad:
removed the console logs that you mentioned. plus found another place with wrong imports

nklomp (approved):
LGTM

PR #84, opened 2023-04-24
uses hashes internally to keep track of VCs instead of ids

### LINKED ISSUE #80: [ios] share flow: send button is always disabled
Hi, I'm trying the share/presentation flow using the `develop` branch in iOS, but I'm stuck in the following screen because the "Send" button is always disabled:

Am I missing something?
Thank you!

nklomp:
fix: Instead of using id values, which are optional to keep track of selections and in summaries of VCs, we now use hashes only, so we can are always able to select VCs even if the original VC does not have an id.

fixes #80


BtencateSphereon (approved):
LGTM

PR #85, opened 2023-04-25
feat: added holder role for VCs

sksadjad:


package.json
```diff
@@ -45,7 +45,7 @@
     "@sphereon/react-native-argon2": "2.0.9",
     "@sphereon/ssi-sdk-contact-manager": "0.9.1-next.103",
     "@sphereon/ssi-sdk-core": "0.9.1-next.103",
-    "@sphereon/ssi-sdk-data-store": "0.9.1-next.103",
+    "@sphereon/ssi-sdk-data-store": "0.9.1-next.107",
```
BtencateSphereon, line 48: need other ssi versions updated as well

src/store/actions/contact.actions.ts
```diff
@@ -97,11 +97,12 @@ const getUserContact = async (): Promise<IContact> => {
     name: userFullName,
     alias: userFullName,
     uri: user.emailAddress,
-    roles: [IdentityRoleEnum.ISSUER],
+    //todo: handle this based on the identities available in the wallet
```
BtencateSphereon, line 100: would be nice to make a ticket for this, so that if we have time we can implement this for v1

package.json
sksadjad: updated all the ssi-sdk versions to `0.9.1-next.107`

src/store/actions/contact.actions.ts
sksadjad: created a ticket and updated

BtencateSphereon (approved):

PR #86, opened 2023-04-25
Wal 513 device event emitter

sanderPostma:
Replaced NativeEventEmitter with DeviceEventEmitter as the former requires native C or Swift code on iOS. Tested on simulator.


BtencateSphereon (approved):
LGTM

PR #88, opened 2023-04-25
Feature/WAL-525

BtencateSphereon:


src/services/qrService.ts
```diff
@@ -51,6 +51,9 @@ import { getContacts } from './contactService'
 import { getOrCreatePrimaryIdentifier } from './identityService'
 import { NativeStackNavigationProp } from '@react-navigation/native-stack'
 import { translateCorrelationIdToName } from '../utils/CredentialUtils'
+import {
+  PresentationDefinitionWithLocation
+} from '@sphereon/did-auth-siop/dist/main/authorization-response'
```
nklomp, line 56: Isn't this exported from the main package? Looking at the code it should be

src/services/qrService.ts
```diff
@@ -205,19 +198,90 @@ const connectSiopV2 = async (args: IQrDataArgs): Promise<void> => {
...
+        filter: [{
+          identities: {
+            identifier: {
+              correlationId: url.hostname,
+            }
+          }
+        }]
+      })
+      if (contacts.length === 1) {
+        const correlationId = request.payload.client_id
```
nklomp, line 222: Since the client_id can be in the OAuth2 part (so the URI), or in the OpenID (requestObject), you should use the the getMergedProperty method. This one checks the requestObject first and then uses the OAuth2 value if not present.

So something like
````typescript
request.authorizationRequest.getMergedProperty('client_id')
````

File: src/services/qrService.ts [hunk 3]
```diff
@@ -205,19 +198,90 @@ const connectSiopV2 = async (args: IQrDataArgs): Promise<void> => {
...
+        const identity: IBasicIdentity = {
+          alias: correlationId,
+          roles: [IdentityRoleEnum.VERIFIER],
+          identifier: {
+            type: CorrelationIdentifierEnum.DID,
```
nklomp, line 227: In theory this could als be a regular URI. So maybe check whether the correlation starts with did: and then use the DID type, otherwise use the URL type

src/services/qrService.ts
```diff
@@ -205,19 +198,90 @@ const connectSiopV2 = async (args: IQrDataArgs): Promise<void> => {
...
+            correlationId
+          }
+        }
+        const hasIdentity = contacts.find((contact: IContact) => contact.identities.some((identity: IIdentity) => identity.identifier.correlationId === correlationId))
+        if (!hasIdentity) {
+          store.dispatch<any>(addIdentity({ contactId: contacts[0].id, identity }))
+        }
+      }
+
+      if (!request.presentationDefinitions || request.presentationDefinitions.length === 0) {
```
nklomp, line 237: Not sure what the impact on later code is. But SIOPv2 and OID4VP are seperate. In other words SIOP doesn't require OID4VP. This means that presentation definitions are optional. 

Probably later code assumes we need definitions, but it would be good to at least already provide some TODO's here that this code should be made optional

src/services/qrService.ts
```diff
@@ -205,19 +198,90 @@ const connectSiopV2 = async (args: IQrDataArgs): Promise<void> => {
...
+        identifier: {
+          correlationId: url.hostname,
+        }
+      }
+    }]
+  }).then((contacts: Array<IContact>) => {
+    if (contacts.length === 0) {
+      args.navigation.navigate(ScreenRoutesEnum.CONTACT_ADD, {
+        name: url.hostname,
+        uri: `${url.protocol}//${url.hostname}`,
```
nklomp, line 266: Missing a colon right?

nklomp (changes requested):
See remarks. The important change is the client_id which can come from 2 places and could also in theory be a non DID value (regular URI)

src/services/qrService.ts
BtencateSphereon: for some reason this is in the protocol

src/services/qrService.ts
BtencateSphereon: ye i think we need to skip the whole required credentials flow in that case?

added comments

src/services/qrService.ts
BtencateSphereon: made some changes

src/services/qrService.ts
BtencateSphereon: ah nice, used that

src/services/qrService.ts
BtencateSphereon: ye it imports by default like this.

changed

nklomp (approved):
LGTM

PR #91, opened 2023-04-25
Fix/wal 242 set translucent

sanderPostma:


App.tsx
```diff
@@ -46,6 +46,25 @@ LogBox.ignoreLogs([
...
+      translucent = false
+    }
+) => {
+  return (
+      <View style={{ height: useSafeAreaInsets().top, backgroundColor }}>
+        <StatusBar
+            animated={true}
+            backgroundColor={backgroundColor}
+            translucent={translucent}
+            barStyle={barStyle} />
```
BtencateSphereon, line 62: i think this is getting a wrong value

BtencateSphereon (changes requested):

App.tsx
sanderPostma: gone

PR #92, opened 2023-04-26
WAL-520: set background color of SSINavigationBar

sanderPostma:
Added a colored view under the SafeAreaView of SSINavigationBar

### DIFF
--- a/src/components/bars/SSINavigationBar/index.tsx
+++ b/src/components/bars/SSINavigationBar/index.tsx
@@ -8,7 +8,7 @@ import {backgrounds, fonts, highLightGradients} from '../../../styles/colors';
 import {
   SSINavigationBarButtonStyled as Button,
   SSINavigationBarContainerStyled as Container,
-  SSINavigationBarSafeAreaContainerStyled
+  SSINavigationBarSafeAreaContainerStyled as SafeAreaContainer
 } from '../../../styles/components';
 import {NavigationBarRoutesEnum} from '../../../types';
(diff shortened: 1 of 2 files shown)


src/components/bars/SSINavigationBar/index.tsx
```diff
@@ -43,45 +43,47 @@ class SSINavigationBar extends PureComponent<BottomTabBarProps, IState> {
 
   render() {
     return !this.state.keyboardVisible ? (
-      <SafeAreaView edges={['bottom']}>
-        <Container>
-          {this.props.state.routes.map((route, index: number) => {
-            const {options} = this.props.descriptors[route.key];
-            const isFocused = this.props.state.index === index;
+        <View style={{backgroundColor: backgrounds.primaryDark}}>
```
BtencateSphereon, line 46: if we go for this option, would like to move this to the styling file of the SSINavigationBar

BtencateSphereon (changes requested):

BtencateSphereon:
I'm not against this solution, but what i think we should do is set the color of the navigation bar in one of the iOS files, like we do for android as that will make sure that if we mess this up somewhere the color is still correct.

Technically with this the color underneath is still white. As already dicussed, i think we should be able to set the color in the AppDelegate.mm for iOS

BtencateSphereon:
Seems that giving that home indicator a color is really complicated on iOS, so i'm fine with the initial approach

BtencateSphereon (approved):

PR #101, opened 2023-04-28
feat/WAL-533

sksadjad:


src/components/pinCodes/SSIPinCode/index.tsx
```diff
@@ -76,7 +76,7 @@ class SSIPinCode extends PureComponent<IProps, IState> {
     const {onVerification} = this.props;
 
     onVerification(value)
-      .then(() => this.setState({retry: 0, pin: ''}))
+      .then(() => this.setState({retry: 0, pin: value}))
```
BtencateSphereon, line 79: the pin in the state should already contain the value. so setting it again is redundant.

BtencateSphereon:
So the ticket was about disabling the input of the pin code. in my opinion, this only "disables" more input, it does not disable the deleting of input. so i wonder if you are still able to remove input while it is executing

BtencateSphereon (changes requested):

src/components/pinCodes/SSIPinCode/index.tsx
sksadjad: if by 'more input' you mean adding text to the pin-code text box, it wasn't the issue, the issue was since we're setting it to an empty value, the empty screen was showed again, and you could type in; and of course it closes immediately

sksadjad:
@BtencateSphereon after we spoke, made a few changes, removed the redundant value assignment as you suggested. also disabled accepting input after we hit the max input length

BtencateSphereon (approved):

PR #111, opened 2023-05-02
fix/WAL-566

sksadjad:
changed the behavior of `getCredentialStatus` to no consider milliseconds

### DIFF
--- a/src/utils/CredentialUtils.ts
+++ b/src/utils/CredentialUtils.ts
@@ -84,10 +84,10 @@ export const isExpired = (value?: string | number): boolean => {
   if (!value) {
     return false;
   }
-  if (typeof value === 'number') {
-    value = value * EPOCH_MILLISECONDS
+  let expirationDate = typeof value === 'string' ? new Date(value).valueOf() : value;
+  if(!isEpochMilli(expirationDate)) {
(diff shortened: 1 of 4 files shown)


src/utils/CredentialUtils.ts
```diff
@@ -46,10 +46,10 @@ export const getOriginalVerifiableCredential = (vc: VerifiableCredential | ICred
   return CredentialMapper.toWrappedVerifiableCredential(vc as OriginalVerifiableCredential).original;
 };
 
-export const getCredentialStatus = (credential: ICredential | VerifiableCredential | ICredentialSummary): CredentialStatusEnum => {
+export const getCredentialStatus = (credential: ICredential | VerifiableCredential | ICredentialSummary, expirationDate: number): CredentialStatusEnum => {
```
BtencateSphereon, line 51: expirationDate should not be a separate param, this should be in the credential. now you could get a missmatch

BtencateSphereon (changes requested):

src/utils/CredentialUtils.ts
sksadjad: handled it within the `isExpired` method instead of passing the value again

BtencateSphereon (approved):

PR #114, opened 2023-05-03
fix/WAL-578

sksadjad:


src/utils/DateUtils.ts
```diff
@@ -7,13 +7,29 @@ export const DATE_FORMAT_OPTIONS = {
...
     hour: 'numeric',
     minute: 'numeric',
     second: 'numeric',
   });
+}
 
-export const toLocalDateString = (date: number): string =>
-  new Date(date * EPOCH_MILLISECONDS).toLocaleDateString(Localization.getLocale(), DATE_FORMAT_OPTIONS);
+export const toLocalDateString = (date: number): string => {
+  date = makeEpochMilli(date)
```
BtencateSphereon, line 21: this is not needed as the result will be only a date?

BtencateSphereon (changes requested):

BtencateSphereon (approved):

PR #115, opened 2023-05-03
fix/WAL-582

BtencateSphereon:


src/providers/authentication/SIOPv2Provider.ts
```diff
@@ -1,8 +1,8 @@
 import {CheckLinkedDomain, VerifiedAuthorizationRequest} from '@sphereon/did-auth-siop';
 import {ConnectionTypeEnum, IDidAuthConfig} from '@sphereon/ssi-sdk-data-store';
-import {OpSession, VerifiablePresentationWithDefinition} from '@sphereon/ssi-sdk-did-auth-siop-authenticator';
-import {getKey} from '@sphereon/ssi-sdk-did-auth-siop-authenticator/dist/session/functions';
-import {VerifiableCredentialsWithDefinition} from '@sphereon/ssi-sdk-did-auth-siop-authenticator/src/types/IDidAuthSiopOpAuthenticator';
+import {OpSession, VerifiableCredentialsWithDefinition, VerifiablePresentationWithDefinition} from '@sphereon/ssi-sdk-did-auth-siop-authenticator';
+import {OID4VP} from '@sphereon/ssi-sdk-did-auth-siop-authenticator/dist/session/OID4VP'; // FIXME we should fix the export of these objects
```
nklomp, line 4: Yeah should be fixed in latest feature branch which has the package name changed as well. That SDK will be released at roughly the same time as the wallet, but it will not be in the first version of the wallet itself

nklomp (approved):
There is a merge conflicht. RESt LGTM

PR #126, opened 2023-05-05
fix/WAL-591

BtencateSphereon:


src/utils/DateUtils.ts
```diff
@@ -18,18 +17,38 @@ export const toLocalDateTimeString = (date: number): string => {
...
-export const isEpochMilli = (epoch: number): boolean => {
-  const epochLength = epoch.toString().length;
-  return epochLength > 10;
+const formatFractionalPart = (date: number): number => {
+  if (date.toString().includes('.')) {
+    const epochParts: Array<string> = date.toString().split('.')
+    const epoch: string = epochParts[0]
+    const fractionalPadded: string = epochParts[1].padEnd(3, '0');
+
+    return Number(`${epoch}${fractionalPadded}`)
```
nklomp, line 50: Hmzz, not sure whether this is always doing the correct thing. Can't we should just drop the fractions altogether here? We are interested in the data anyway, not the exact time.

We have code to determine whether it is milliseconds or seconds. So the fraction could be about milliseconds or seconds as well. I think the code assumes seconds for the fractional number and then creates milliseconds for the result because it simply adds them to the number

nklomp (commented):
See remarks. Not sure about the fractional handling tbh

src/utils/DateUtils.ts
BtencateSphereon: Seems to have the same effect when the fractional part is removed. So I removed it.

PR #128, opened 2023-05-05
feature/WAL-586

sksadjad:


src/services/qrService.ts
```diff
@@ -380,7 +381,7 @@ const connectOpenId4VcIssuance = async (args: IQrDataArgs): Promise<void> => {
     }).then((contacts: Array<IContact>) => {
       if (contacts.length === 0) {
         args.navigation.navigate(ScreenRoutesEnum.CONTACT_ADD, {
-          name: url.hostname,
+          name,
```
nklomp, line 384: name || url.hostname

nklomp (commented):
should fallback to url as mentioned

src/services/qrService.ts
sksadjad: I'm doing exactly that a few lines above 370

nklomp: Oh okay, because of the long line in Github didn't see that

nklomp (approved):

PR #131, opened 2023-05-08
fixes/WAL-597_qr-crash

sanderPostma:
PS. also ran into 
  StatusBar.setTranslucent(true);
  StatusBar.setBackgroundColor('transparent');
instances, they are now only executed for Android.


nklomp (approved):
LGTM

PR #132, opened 2023-05-09
feature/WAL-521

sksadjad:


src/components/containers/SSIButtonsContainer/index.tsx
```diff
@@ -41,11 +42,11 @@ class SSIButtonsContainer extends PureComponent<Props, IState> {
   };
 
   render() {
-    const {primaryButton, secondaryButton} = this.props;
+    const {primaryButton, secondaryButton, style} = this.props;
     const {keyboardVisible} = this.state;
 
     return (
-      <ButtonContainer style={{marginBottom: keyboardVisible ? 18 : 36}}>
+        <ButtonContainer style={{marginBottom: keyboardVisible ? 18 : 36, ...(style && style)}}>
```
BtencateSphereon, line 49: style={{...style, marginBottom: keyboardVisible ? 18 : 36}}

you can just deconstruct in this case

src/components/views/SSITabView/index.tsx
```diff
@@ -7,17 +8,18 @@ import {ITabRoute, ITabViewRoute} from '../../../types';
...
 export interface IProps {
+  style?: ViewStyle;
   routes: Array<ITabViewRoute>;
 }
 
 const SSITabView: FC<IProps> = (props: IProps): JSX.Element => {
   const [index, setIndex] = React.useState(0);
   const [routes] = React.useState(props.routes);
-
+  const style = props.style;
```
BtencateSphereon, line 18: i rather deconstruct props where possible:

const { style } = props

then you can use style where you want. you can then also try to at routes to it, but it shares the name of the state

src/components/views/SSITabView/index.tsx
```diff
@@ -7,17 +8,18 @@ import {ITabRoute, ITabViewRoute} from '../../../types';
...
   const sceneMap = props.routes.reduce((a: Record<string, any>, v: ITabViewRoute) => ({...a, [v.key]: v.content}), {});
 
   return (
-    <Container>
+    <Container style={{...(style && style)}}>
```
BtencateSphereon, line 22: is this needed? the tabview just hold some routes, these routes should determine which color of background they need

src/screens/SSICredentialDetailsScreen/index.tsx
```diff
@@ -62,7 +63,7 @@ const SSICredentialDetailsScreen: FC<Props> = (props: Props): JSX.Element => {
         <SSITabView routes={routes} />
         {/* TODO we use this 2 button structure a lot, we should make a component out of it */}
         {(primaryAction || secondaryAction) && (
-          <ButtonContainer>
+          <ButtonContainer style={{ backgroundColor: backgrounds.secondaryDark }}>
```
BtencateSphereon, line 66: we could replace this whole view with the SSIButtonContainer component. if that is to much work, move the styling here to the correct styling component

src/screens/SSISummaryScreen/index.tsx
```diff
@@ -65,8 +66,8 @@ class SSIOnboardingSummaryScreen extends PureComponent<IProps> {
     ];
 
     return (
-      <Container>
-        <SSITabView routes={routes} />
+      <Container style={{ backgroundColor: backgrounds.secondaryDark }}>
```
BtencateSphereon, line 69: same comment here, what is the reason for adding this? and can the routes not decide which color they want? same for the styling below this

BtencateSphereon (changes requested):

src/components/views/SSITabView/index.tsx
sksadjad: removed this styling

src/components/views/SSITabView/index.tsx
sksadjad: remvoed this styling

src/components/containers/SSIButtonsContainer/index.tsx
sksadjad: removed this styling

src/screens/SSISummaryScreen/index.tsx
```diff
@@ -65,8 +66,8 @@ class SSIOnboardingSummaryScreen extends PureComponent<IProps> {
     ];
 
     return (
-      <Container>
-        <SSITabView routes={routes} />
+      <Container style={{backgroundColor:backgrounds.secondaryDark}}>
```
BtencateSphereon, line 69: Create a new styling component for this container specifically for this screen with the background color

sksadjad: created a new component for `SSISummaryScreen` container

BtencateSphereon (approved):

PR #135, opened 2023-05-10
[WAL-540] allow edit while credential issuance

hrehman-sphereon:


src/store/actions/contact.actions.ts
```diff
@@ -50,6 +54,29 @@ export const createContact = (args: ICreateContactArgs): ThunkAction<Promise<voi
   };
 };
 
+export const updateContact = (args: IUpdateContactArgs): ThunkAction<Promise<void>, RootState, unknown, Action> => {
+  return async (dispatch: ThunkDispatch<RootState, unknown, Action>) => {
+    dispatch({type: CONTACTS_LOADING});
+    editContact(args)
+    .then(()=> {
```
BtencateSphereon, line 61: please refactor this  to a promise chain that is not nested with only 1 catch at the end. this is unreadable

src/types/store/contact.action.types.ts
```diff
@@ -41,6 +45,15 @@ interface ICreateContactFailedAction {
   type: CREATE_CONTACT_FAILED;
 }
 
+interface IUpdateContactSuccessAction {
+  type: UPDATE_CONTACT_SUCCESS;
+  payload: Array<IContact>;
```
BtencateSphereon, line 50: should only be 1 contact

src/store/reducers/contact.reducer.ts
```diff
@@ -53,10 +56,23 @@ const contactReducer = (state: IContactState = initialState, action: ContactActi
         loading: false,
       };
     }
+    case UPDATE_CONTACT_SUCCESS: {
+      return {
+        ...state,
+        contacts: action.payload,
```
BtencateSphereon, line 62: missing the sort?

src/services/qrService.ts
```diff
@@ -456,12 +456,6 @@ const connectOpenId4VcIssuance = async (args: IQrDataArgs): Promise<void> => {
           },
         },
       });
-      // TODO WAL-540 do not filter CONTACT_ADD, this route should support edit contact
```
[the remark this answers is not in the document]
hrehman-sphereon: reverted. Upon realizing removed was doing way more than where I left it

src/store/actions/contact.actions.ts
```diff
@@ -50,6 +54,29 @@ export const createContact = (args: ICreateContactArgs): ThunkAction<Promise<voi
...
+      getUserContact()
+        .then((userContact: IContact) => {
+          getContactsFromStorage()
+          .then(async (contacts: Array<IContact>) => {
+            dispatch({
+              type: UPDATE_CONTACT_SUCCESS,
+              payload: [
```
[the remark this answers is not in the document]
hrehman-sphereon: The reason to do it this way was that I found another route. And that route was to go to the contact screen before accepting VC. I was not seeing the right amount of contacts on the contact list. I will try again with one contact returned from update.

src/utils/ContactUtil.ts
```diff
@@ -0,0 +1,5 @@
+import {IContact} from "@sphereon/ssi-sdk-data-store";
+
+export function compareContact(contact1: IContact, contact2: IContact) {
```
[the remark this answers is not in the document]
hrehman-sphereon: changed it.

src/store/actions/contact.actions.ts
hrehman-sphereon: made to return single contact

src/types/store/contact.action.types.ts
hrehman-sphereon: https://github.com/Sphereon-Opensource/ssi-mobile-wallet/pull/135#discussion_r1191163081

Fixed it as per review comment.

src/screens/SSIContactAddScreen/index.tsx
```diff
@@ -54,26 +54,37 @@ class SSIContactAddScreen extends PureComponent<IProps, IState> {
...
-          name,
-          alias: contactAlias.trim(),
-          uri,
-          identities,
-        });
-        onCreate();
+    this.onValidate(this.state.contactAlias)
+      .then(async () => {
+        await this.upsert();
+        await this.props.route.params.onCreate();
```
BtencateSphereon, line 62: i rather deconstruct these things. so

const { onCreate } = this.props.route.params

and then call await onCreate()

keeps the code a lot cleaner and more readable if we get more

src/services/qrService.ts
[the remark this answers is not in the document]
BtencateSphereon: the comments pretty mush already stated what needed to be removed

src/store/actions/contact.actions.ts
```diff
@@ -50,6 +54,20 @@ export const createContact = (args: ICreateContactArgs): ThunkAction<Promise<voi
   };
 };
 
+export const updateContact = (args: IUpdateContactArgs): ThunkAction<Promise<void>, RootState, unknown, Action> => {
+  return async (dispatch: ThunkDispatch<RootState, unknown, Action>) => {
+    dispatch({type: CONTACTS_LOADING});
+    editContact(args)
+      .then(() =>
```
BtencateSphereon, line 61: i would give this a result and then use that as the payload, now there could be a difference between the args and the actual updated contact

src/utils/SortUtil.ts
```diff
@@ -0,0 +1,14 @@
+export enum SortOrder {
```
BtencateSphereon, line 1: keep the file names consistant. so sortUtils

src/utils/SortUtil.ts
```diff
@@ -0,0 +1,14 @@
...
+  ASC = 'asc',
```
BtencateSphereon, line 2: not too important,

but introducing the concept of sorting is fine to me, but why not then add to the other places as well? credentials? identities?

now only contacts are sorted, which does not make much sense to me if you spent time on getting sorting implemented

can we atleast put some comments on the other places and create a ticket for it?

src/store/reducers/contact.reducer.ts
BtencateSphereon: now the assignment is incorrect. you are assigning a single contact to an array

src/store/reducers/contact.reducer.ts
```diff
@@ -11,15 +11,24 @@ import {
...
 import {IContactState} from '../../types/store/contact.types';
+import {sortBy, SortOrder} from '../../utils/SortUtil';
 
 const initialState: IContactState = {
   loading: false,
   contacts: [],
 };
 
 const contactReducer = (state: IContactState = initialState, action: ContactActionTypes): IContactState => {
+  if ('payload' in action) {
```
BtencateSphereon, line 26: this will not work as some payloads are singular, so you first need to add the payload (if it's a contact) to the contacts and then assign it to the state

BtencateSphereon: ALso what happends if the payload is not a contact or contacts?

src/screens/SSIContactAddScreen/index.tsx
hrehman-sphereon: changed.

src/screens/SSIContactAddScreen/index.tsx
```diff
@@ -54,26 +54,37 @@ class SSIContactAddScreen extends PureComponent<IProps, IState> {
...
 
+  private async upsert() {
+    const {identities, name, uri} = this.props.route.params;
+    const {contactAlias} = this.state;
+
+    const contacts = await getContacts({filter: [{name: name}]});
+    if (contacts.length !== 0) {
+      const contactToUpdate: IUpdateContactArgs = {contact: contacts[0]};
+      contactToUpdate.contact.alias = contactAlias;
+      this.props.updateContact(contactToUpdate);
```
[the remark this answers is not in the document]
hrehman-sphereon: changed.

src/store/actions/contact.actions.ts
hrehman-sphereon: channged

src/store/reducers/contact.reducer.ts
hrehman-sphereon: I think code has been miss-read here. Please have a look again. 
I am altering the param. And I have seen it work at least for add and update contact.
I think this is centeral enough place and there is a check whether the payload is Array or not.
I will test other cases as well.

src/store/reducers/contact.reducer.ts
hrehman-sphereon: I don't like this about React. A compilation error can be hidden while the application seem to work. I will test again.

src/utils/SortUtil.ts
hrehman-sphereon: May be it's miss read ... the SortOrder is the enum ... and has the correct capitalization.
The file name seem to have correct capitalization. 

And the name of the function sortBy has the correct capitalization.

src/utils/SortUtil.ts
hrehman-sphereon: Other places are not part of this ticket. I would request to not expand. and let this ticket go towards closure. 

I will create a ticket just as a reminder.
https://sphereon.atlassian.net/browse/WAL-605

src/store/reducers/contact.reducer.ts
hrehman-sphereon: fixed

src/store/reducers/contact.reducer.ts
hrehman-sphereon: delete also works.
Get also works

src/utils/SortUtil.ts
BtencateSphereon: all the util files are called XUtils. this one is called XUtil

src/utils/SortUtil.ts
BtencateSphereon: The sorting for contacts was also not part of the ticket. lets add some todo's on the other places that references the ticket you made

hrehman-sphereon: I was trying to avoid even the references in the code. But I will shortly do add those.

There were 2 rules.
1- Don't change anything that is not part of the ticket.
2- If the ticket inroduces a bug it will be fixed in the same ticket. Since, the changes in the Contact reducer were made and it was seen that the over all combination of running different reducers resulted in different ordering. Hence, this change was made. 

I opted the 2nd rule.

Also, in previous ticket(s) the "other" developer also makes such small changes. As the code-author wills.

nklomp: Yeah, but either do it right then or don't do it

nklomp: So do it all, or create a ticket. Not do partially and be done with it

hrehman-sphereon: Then there will be comments about intorudcing the sorting issue and not fixing it. 
So there are 3 conflicting rules.
Winner is the most persistent disagreeable one.

src/store/reducers/contact.reducer.ts
BtencateSphereon: the problem is that you sort if it is an array, but then add the singular contact to the sorted array, which i think is then not sorted properly anymore

BtencateSphereon (approved):

PR #136, opened 2023-05-11
[draft] temp/WAL-521_experimental

sksadjad:


sksadjad:
this was just to showcase the possibilities we have, after some discussions around it, we don't need this pr anymore

PR #138, opened 2023-05-11
fix/WAL-599

BtencateSphereon:


android/app/src/main/java/com/sphereon/ssi/wallet/MainActivity.java
```diff
@@ -71,5 +123,54 @@ public void invokeDefaultOnBackPressed() {
     super.invokeDefaultOnBackPressed();
   }
 
+  @Override
+  protected void onStop() {
+    super.onStop();
```
nklomp, line 128: Are you sure you want to call this first and not last for instance?

android/app/src/main/java/com/sphereon/ssi/wallet/MainActivity.java
```diff
@@ -71,5 +123,54 @@ public void invokeDefaultOnBackPressed() {
...
+    }
+  }
+
+  private boolean isAppMovingToBackground() {
+    ActivityManager activityManager = (ActivityManager) getSystemService(Context.ACTIVITY_SERVICE);
+    if (activityManager == null) {
+      return false;
+    }
+
+    List<ActivityManager.RunningTaskInfo> tasks = activityManager.getRunningTasks(1);
```
nklomp, line 140: See https://developer.android.com/reference/android/app/ActivityManager#getRunningTasks(int). 

It is deprecated as of 21 and will be removed (or is even removed in newer versions)
It is also only meant for debug purposes or task applications according to the link

nklomp (changes requested):
See remarks.

The getTasks call is deprecated and it is explicitly mentioned in the docs to not use it.

Also where would the iOS handling be? This needs to go in sync with the iOS implementation when we want to replace current locking implementation

android/app/src/main/java/com/sphereon/ssi/wallet/MainActivity.java
BtencateSphereon: After some research, I think the best practice is as follows:

Methods you override that are part of component creation (onCreate(), onStart(), onResume(), etc.), you should chain to the superclass as the first statement. This way android can do it's think before calling custom code that might depend on this.

Methods you override that are part of component destruction (onPause(), onStop(), onDestroy(), etc.), you should do your work first and chain to the superclass as the last thing. This way android cannot clean things up that the custom code might rely on

so in this case i moved the super call for the onStop and onDestroy

BtencateSphereon: this has completely been refactored

android/app/src/main/java/com/sphereon/ssi/wallet/MainActivity.java
BtencateSphereon: this has completely been refactored

nklomp (approved):
LGTM

PR #141, opened 2023-05-16
feature/WAL-606

BtencateSphereon:


src/components/bars/SSIHeaderBar/index.tsx
```diff
@@ -62,6 +65,24 @@ const SSIHeaderBar: FC<Props> = (props: Props): JSX.Element => {
     dispatch<any>(logout());
   };
 
+  const onDeleteWallet = async (): Promise<void> => {
+    setShowProfileMenu(false);
+    const activeUser: IUser = store.getState().user.activeUser!;
+
+    props.navigation.navigate(MainRoutesEnum.POPUP_MODAL, {
+      title: translate('profile_delete_wallet_action_title'),
+      details: format(translate('profile_delete_wallet_action_subtitle'), `${activeUser.firstName} ${activeUser.lastName}`),
```
nklomp, line 74: It would be nicer if we adjust the translate method to accept a vararg for variables

src/services/userService.ts
```diff
@@ -48,3 +48,7 @@ export const updateUser = async (args: IUser): Promise<IUser> => {
     })
     .catch((error: Error) => Promise.reject(Error(`Unable to update user. Error: ${error}`)));
 };
+
+export const deleteUser = async (userId: string): Promise<void> => {
+  removeUser(userId).catch((error: Error) => Promise.reject(Error(`Unable to delete user with id: ${userId}. Error: ${error}`)));
```
nklomp, line 53: translate missing

src/services/storageService.ts
```diff
@@ -66,3 +71,26 @@ export const getUsers = async (): Promise<Map<string, IUser>> => {
...
+      JSON.parse(result).forEach(([key, value]: [string, IUser]) => {
+        users.set(key, value);
+      });
+
+      debug(`deleteUser(${userId}) deleting user`);
+      users.delete(userId);
+
+      AsyncStorage.setItem(STORAGE_USERS_KEY, JSON.stringify(Array.from(users)));
+    })
+    .catch((error: Error) => Promise.reject(new Error(`Unable to set value for key: ${STORAGE_PIN_KEY}. ${error.message}`)));
```
nklomp, line 95: translate missing

nklomp (changes requested):
Some small remarks about translations (errors we generate the strings for).
Also it would make sense to add varargs to the translate function so we can always pas in arguments when a message contains variables

src/components/bars/SSIHeaderBar/index.tsx
BtencateSphereon: refactored it as i18n could already do this

src/services/userService.ts
BtencateSphereon: we are not showing this to a user, and we are not translating any other error. do you want me to add this?

src/services/storageService.ts
BtencateSphereon: we are not showing this to a user, and we are not translating any other error. do you want me to add this?

nklomp (approved):
LGTM. Will merge

PR #145, opened 2023-06-13
branding / issuance fixes

BtencateSphereon:


src/providers/credential/OpenId4VcIssuanceProvider.ts
```diff
@@ -255,16 +254,20 @@ class OpenId4VcIssuanceProvider {
...
             (metadata.display ?? []).map(
-              async (display: CredentialDisplay): Promise<IBasicCredentialLocaleBranding> =>
+              async (display: MetadataDisplay): Promise<IBasicCredentialLocaleBranding> =>
                 await ibCredentialLocaleBrandingFrom({localeBranding: await credentialLocaleBrandingFrom(display)}),
             ),
           );
 
+          const credentialTypes: Array<string> = metadata.types.length > 1
+            ? metadata.types.filter((type: string) => type !== 'VerifiableCredential')
+            : metadata.types
```
nklomp, line 267: Probably wise to also check for an empty array, as then the result should be an array with only VerifiableCredential in there

nklomp (changes requested):
See remark about a fallback for when the array is empty

src/providers/credential/OpenId4VcIssuanceProvider.ts
BtencateSphereon: added a fix for that

nklomp (approved):

PR #150, opened 2023-06-27
fix/WAL-609

BtencateSphereon:


src/services/qrService.ts
```diff
@@ -68,373 +57,379 @@ import {
...
+      screen: ScreenRoutesEnum.CREDENTIALS_REQUIRED,
+      params: {
+        verifier: translateCorrelationIdToName(url.hostname),
+        // TODO currently only supporting 1 presentation definition
+        presentationDefinition: presentationDefinitionWithLocation.definition,
+        format,
+        subjectSyntaxTypesSupported,
+        onDecline: async (): Promise<void> => {
+          // TODO WAL-541 fix navigation hierarchy
+          args.navigation.navigate(ScreenRoutesEnum.QR_READER, {});
```
nklomp, line 326: Does this mean that when the user is using a deeplink and declines, they will go to the QR code reader?

IMO this should go to the list credentials/home screen anyway. Even in a non-deeplink use case: If I scan a QR code then decline the credential, I would not expect to get back to the QR code scanner. The QR either is gone now, or it is still the same QR. Going to the overview screen simply makes more sense and is more natural

src/services/qrService.ts
```diff
@@ -476,13 +471,13 @@ const connectOpenId4VcIssuance = async (args: IQrDataArgs): Promise<void> => {
                 },
               },
             ],
-            onDecline: async () => {
+            onDecline: async (): Promise<void> => {
               // TODO WAL-541 fix navigation hierarchy
-              args.navigation.navigate(ScreenRoutesEnum.QR_READER, {})
+              args.navigation.navigate(ScreenRoutesEnum.QR_READER, {});
```
nklomp, line 476: Does this mean that when the user is using a deeplink and declines, they will go to the QR code reader?

IMO this should go to the list credentials/home screen anyway. Even in a non-deeplink use case: If I scan a QR code then decline the credential, I would not expect to get back to the QR code scanner. The QR either is gone now, or it is still the same QR. Going to the overview screen simply makes more sense and is more natural

nklomp (commented):
See remarks about decline.

Also we need to fix the formating/prettier stuff. The PR itself is actually pretty small, but because of all the formatting changes it is very hard to disect

src/services/qrService.ts
BtencateSphereon: changed navigation to go to credentials overview

src/services/qrService.ts
BtencateSphereon: changed navigation to go to credentials overview

nklomp (approved):
LGTM

PR #157, opened 2023-08-11
FIX: provide correct did:web resolver configuration

linasi:
Hello,

SSI Wallet fails verify credentials when issuer uses `did:web` method.
Wallet throws `unsupportedDIDMethod` error that originates from the instance of the `Resolver` when mapping is not found.

`web-did-resolver` resolver is not provided correctly.


BtencateSphereon (approved):
Hi @linasi, thanks for spotting this and providing a fix for it. PR approved 👍

BtencateSphereon:
Hi @linasi, we have a few delays with a new release of the app. We are working on getting one asap :)

PR #169, opened 2023-10-20
Fix required creds screen

sanderPostma:
When not having a valid credential in your wallet, it is showing all available credentials


sanderPostma:
Not a good solution

PR #170, opened 2023-10-20
Fix required creds screen

sanderPostma:
When not having a valid credential in your wallet, it is showing all available credentials

### DIFF
--- a/src/screens/SSICredentialsRequiredScreen/index.tsx
+++ b/src/screens/SSICredentialsRequiredScreen/index.tsx
@@ -161,17 +161,19 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
         ? undefined
         : () => onItemPress(itemInfo.item.id, availableCredentials.get(itemInfo.item.id)!, itemInfo.item.purpose);
 
-    const checkIsMatching = (itemInfo: any, selectedCredentials: any, pex: any, getOriginalVerifiableCredential: any) => {
+    const checkIsMatching = (itemInfo: ListRenderItemInfo<InputDescriptorV1 | InputDescriptorV2>,
+                             selectedCredentials: Map<string, Array<UniqueVerifiableCredential>>) => {
       if (!selectedCredentials.has(itemInfo.item.id)) {
         return false;
       }
 
-      const credentials = selectedCredentials.get(itemInfo.item.id)!.map((uniqueVC: any) => getOriginalVerifiableCredential(uniqueVC.verifiableCredential));
+      const credentials = selectedCredentials.get(itemInfo.item.id)!.map((uniqueVC: UniqueVerifiableCredential) =>
+        getOriginalVerifiableCredential(uniqueVC.verifiableCredential));
 
       return pex.evaluateCredentials(
         {
           id: itemInfo.item.id,
-          input_descriptors: [itemInfo.item],
+          input_descriptors: [itemInfo.item]
         },
         credentials
       ).areRequiredCredentialsPresent === Status.INFO;
@@ -184,7 +186,7 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
         purpose={itemInfo.item.purpose}
         available={availableCredentials.has(itemInfo.item.id) ? availableCredentials.get(itemInfo.item.id)! : undefined}
         selected={selectedCredentials.has(itemInfo.item.id) ? selectedCredentials.get(itemInfo.item.id)! : []}
-        isMatching={checkIsMatching(itemInfo, selectedCredentials, pex, getOriginalVerifiableCredential)}
+        isMatching={checkIsMatching(itemInfo, selectedCredentials)}
         listIndex={itemInfo.index}
         onPress={onPress}
       />
--- a/src/screens/SSICredentialsRequiredScreen/index.tsx
+++ b/src/screens/SSICredentialsRequiredScreen/index.tsx
@@ -53,17 +53,17 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
           restrictToFormats: format,
           restrictToDIDMethods: subjectSyntaxTypesSupported,
         });
-        if (selectResult.areRequiredCredentialsPresent === "error") {
+        if (selectResult.areRequiredCredentialsPresent === Status.ERROR) {
           console.debug('pex.selectFrom returned errors:\n', JSON.stringify(selectResult.errors));
         }
         const matchedVCs: Array<UniqueVerifiableCredential> = selectResult.matches && selectResult.verifiableCredential
           ? selectResult.matches.map((match: SubmissionRequirementMatch) => {
-            const matchedVC = JSONPath.query(selectResult, match.vc_path[0])
-            if (matchedVC?.length) {
+            const matchedVC = JSONPath.query(selectResult, match.vc_path[0]) // TODO Can we have multiple vc_path elements for a single match?
+            if (matchedVC?.length > 0) {
               return getMatchingUniqueVerifiableCredential(uniqueVCs, matchedVC[0]);
             }
           })
-          .filter((matchedVC): matchedVC is UniqueVerifiableCredential => !!matchedVC) // filter out the undefined (should not happen)
+          .filter((matchedVC : UniqueVerifiableCredential | undefined): matchedVC is UniqueVerifiableCredential => !!matchedVC) // filter out the undefined (should not happen)
           : [];
         availableVCs.set(inputDescriptor.id, matchedVCs);
       });
--- a/src/screens/SSICredentialsRequiredScreen/index.tsx
+++ b/src/screens/SSICredentialsRequiredScreen/index.tsx
@@ -9,6 +9,7 @@ import React, {FC, useEffect, useState} from 'react';
 import {ListRenderItemInfo} from 'react-native';
 import {SwipeListView} from 'react-native-swipe-list-view';
 
+
 import {OVERVIEW_INITIAL_NUMBER_TO_RENDER} from '../../@config/constants';
 import {ibGetCredentialBranding} from '../../agent';
 import SSIButtonsContainer from '../../components/containers/SSIButtonsContainer';
@@ -23,6 +24,8 @@ import {
 import {ScreenRoutesEnum, StackParamList} from '../../types';
 import {getMatchingUniqueVerifiableCredential, getOriginalVerifiableCredential} from '../../utils/CredentialUtils';
 import {toCredentialSummary} from '../../utils/mappers/credential/CredentialMapper';
+import { SubmissionRequirementMatch } from "@sphereon/pex/dist/main/lib/evaluation/core/submissionRequirementMatch";
+import { JSONPath } from "@astronautlabs/jsonpath/src/jsonpath";
 
 type Props = NativeStackScreenProps<StackParamList, ScreenRoutesEnum.CREDENTIALS_REQUIRED>;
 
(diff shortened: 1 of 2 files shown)


src/screens/SSICredentialsRequiredScreen/index.tsx
```diff
@@ -23,6 +24,8 @@ import {
 import {ScreenRoutesEnum, StackParamList} from '../../types';
 import {getMatchingUniqueVerifiableCredential, getOriginalVerifiableCredential} from '../../utils/CredentialUtils';
 import {toCredentialSummary} from '../../utils/mappers/credential/CredentialMapper';
+import { SubmissionRequirementMatch } from "@sphereon/pex/dist/main/lib/evaluation/core/submissionRequirementMatch";
```
nklomp, line 27: What's up with this import?

src/screens/SSICredentialsRequiredScreen/index.tsx
```diff
@@ -50,10 +53,17 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
           restrictToFormats: format,
           restrictToDIDMethods: subjectSyntaxTypesSupported,
         });
-        const matchedVCs: Array<UniqueVerifiableCredential> = selectResult.verifiableCredential
-          ? selectResult.verifiableCredential
-              .map((matchedVC: IVerifiableCredential) => getMatchingUniqueVerifiableCredential(uniqueVCs, matchedVC))
-              .filter((matchedVC): matchedVC is UniqueVerifiableCredential => !!matchedVC) // filter out the undefined (should not happen)
+        if (selectResult.areRequiredCredentialsPresent === "error") {
+          console.info('pex.selectFrom returned errors:\n', JSON.stringify(selectResult.errors));
```
nklomp, line 57: Debug package call or remove

src/screens/SSICredentialsRequiredScreen/index.tsx
```diff
@@ -151,24 +161,30 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
         ? undefined
         : () => onItemPress(itemInfo.item.id, availableCredentials.get(itemInfo.item.id)!, itemInfo.item.purpose);
 
+    const checkIsMatching = (itemInfo: any, selectedCredentials: any, pex: any, getOriginalVerifiableCredential: any) => {
+      if (!selectedCredentials.has(itemInfo.item.id)) {
+        return false;
+      }
+
+      const credentials = selectedCredentials.get(itemInfo.item.id)!.map((uniqueVC: any) => getOriginalVerifiableCredential(uniqueVC.verifiableCredential));
```
nklomp, line 169: Don't introduce any

src/screens/SSICredentialsRequiredScreen/index.tsx
```diff
@@ -151,24 +161,30 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
         ? undefined
         : () => onItemPress(itemInfo.item.id, availableCredentials.get(itemInfo.item.id)!, itemInfo.item.purpose);
 
+    const checkIsMatching = (itemInfo: any, selectedCredentials: any, pex: any, getOriginalVerifiableCredential: any) => {
```
nklomp, line 164: Don't introduce any. We have types for these

nklomp (changes requested):
See remarks

nklomp:
Also last commit changing the guard on the length can have undesired effects. Why did you change a guard that tested a length to be bigger than 0 to only checking for a length, which could be zero and then the next line taking the first element from the array...

sanderPostma:
> Also last commit changing the guard on the length can have undesired effects. Why did you change a guard that tested a length to be bigger than 0 to only checking for a length, which could be zero and then the next line taking the first element from the array...

image

src/screens/SSICredentialsRequiredScreen/index.tsx
sanderPostma: It's coming from selectResult.matches

src/screens/SSICredentialsRequiredScreen/index.tsx
```diff
@@ -50,10 +53,17 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
...
-          ? selectResult.verifiableCredential
-              .map((matchedVC: IVerifiableCredential) => getMatchingUniqueVerifiableCredential(uniqueVCs, matchedVC))
-              .filter((matchedVC): matchedVC is UniqueVerifiableCredential => !!matchedVC) // filter out the undefined (should not happen)
+        if (selectResult.areRequiredCredentialsPresent === "error") {
+          console.info('pex.selectFrom returned errors:\n', JSON.stringify(selectResult.errors));
+        }
+        const matchedVCs: Array<UniqueVerifiableCredential> = selectResult.matches && selectResult.verifiableCredential
+          ? selectResult.matches.map((match: SubmissionRequirementMatch) => {
+            const matchedVC = JSONPath.query(selectResult, match.vc_path[0])
+            if(matchedVC && matchedVC.length) {
```
BtencateSphereon, line 62: can be shortened by

if (matchedVC?.lenght) {

src/screens/SSICredentialsRequiredScreen/index.tsx
sanderPostma: fixed

src/screens/SSICredentialsRequiredScreen/index.tsx
BtencateSphereon: can you not use?

import { SubmissionRequirementMatch } from "@sphereon/pex";

src/screens/SSICredentialsRequiredScreen/index.tsx
sanderPostma: That was bad yes. Limited to GPT 3.5 atm and forgot to check what it wrote.

src/screens/SSICredentialsRequiredScreen/index.tsx
```diff
@@ -50,10 +53,17 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
           restrictToFormats: format,
           restrictToDIDMethods: subjectSyntaxTypesSupported,
         });
-        const matchedVCs: Array<UniqueVerifiableCredential> = selectResult.verifiableCredential
-          ? selectResult.verifiableCredential
-              .map((matchedVC: IVerifiableCredential) => getMatchingUniqueVerifiableCredential(uniqueVCs, matchedVC))
-              .filter((matchedVC): matchedVC is UniqueVerifiableCredential => !!matchedVC) // filter out the undefined (should not happen)
+        if (selectResult.areRequiredCredentialsPresent === "error") {
```
BtencateSphereon, line 56: should we not check against Status.ERROR?

nklomp:
Reality ugly and less readable so don't please

src/screens/SSICredentialsRequiredScreen/index.tsx
sanderPostma: done

src/screens/SSICredentialsRequiredScreen/index.tsx
```diff
@@ -50,10 +53,17 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
...
+          console.info('pex.selectFrom returned errors:\n', JSON.stringify(selectResult.errors));
+        }
+        const matchedVCs: Array<UniqueVerifiableCredential> = selectResult.matches && selectResult.verifiableCredential
```
BtencateSphereon, line 59: why are we checking selectResult.verifiableCredential and not using that? if it has matches it should be ok right? selectResult.matches

src/screens/SSICredentialsRequiredScreen/index.tsx
nklomp: Uhm in our code you really should not use chatgpt for these types of fixes, tickets

src/screens/SSICredentialsRequiredScreen/index.tsx
```diff
@@ -50,10 +53,17 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
...
+          ? selectResult.matches.map((match: SubmissionRequirementMatch) => {
+            const matchedVC = JSONPath.query(selectResult, match.vc_path[0])
```
BtencateSphereon, line 61: can we add some comments explaining why we only pick the first one?

src/screens/SSICredentialsRequiredScreen/index.tsx
```diff
@@ -50,10 +53,17 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
...
+            if(matchedVC && matchedVC.length) {
+              return getMatchingUniqueVerifiableCredential(uniqueVCs, matchedVC[0]);
+            }
+          })
+          .filter((matchedVC): matchedVC is UniqueVerifiableCredential => !!matchedVC) // filter out the undefined (should not happen)
```
BtencateSphereon, line 66: missing a type

src/screens/SSICredentialsRequiredScreen/index.tsx
nklomp: Yeah, the question obviously wasn't about why there is an import, it was about why the package wasn't imported

src/screens/SSICredentialsRequiredScreen/index.tsx
sanderPostma: Ah it has an enum. So yes, done

src/screens/SSICredentialsRequiredScreen/index.tsx
sanderPostma: The response is like this:
```
{
	"errors": [],
	"matches": [
		{
			"name": "Account",
			"rule": "all",
			"vc_path": [
				"$.verifiableCredential[0]"
			]
		}
	],
	"areRequiredCredentialsPresent": "info",
	"verifiableCredential": [
		"eyJh*snip*"
	],
	"warnings": []
}
```

nklomp: Yes and better be too defensive in cases like this

src/screens/SSICredentialsRequiredScreen/index.tsx
BtencateSphereon: @nklomp technically matchedVC?.lenght is only true for 1 orr more. but i do agree that using > 0 is more readable

src/screens/SSICredentialsRequiredScreen/index.tsx
sanderPostma: Can we have multiple vc_path elements for a single match?

nklomp: Not entirely sure from the top of my head   but given the json path would allow for array results it stems to reason that multiple matches should be allowed yes and that all should be returned

src/screens/SSICredentialsRequiredScreen/index.tsx
```diff
@@ -50,10 +53,17 @@ const SSICredentialsRequiredScreen: FC<Props> = (props: Props): JSX.Element => {
...
-              .map((matchedVC: IVerifiableCredential) => getMatchingUniqueVerifiableCredential(uniqueVCs, matchedVC))
-              .filter((matchedVC): matchedVC is UniqueVerifiableCredential => !!matchedVC) // filter out the undefined (should not happen)
+        if (selectResult.areRequiredCredentialsPresent === "error") {
+          console.info('pex.selectFrom returned errors:\n', JSON.stringify(selectResult.errors));
+        }
+        const matchedVCs: Array<UniqueVerifiableCredential> = selectResult.matches && selectResult.verifiableCredential
+          ? selectResult.matches.map((match: SubmissionRequirementMatch) => {
+            const matchedVC = JSONPath.query(selectResult, match.vc_path[0])
+            if(matchedVC && matchedVC.length) {
+              return getMatchingUniqueVerifiableCredential(uniqueVCs, matchedVC[0]);
```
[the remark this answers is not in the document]
sanderPostma: // TODO Can we have multiple vc_path elements for a single match?

src/screens/SSICredentialsRequiredScreen/index.tsx
sanderPostma: fixed

src/screens/SSICredentialsRequiredScreen/index.tsx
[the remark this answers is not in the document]
BtencateSphereon: tbh, i haven't worked with pex for a while and cannot remember it all. So i do not think i can answer this atm without looking into it.

src/screens/SSICredentialsRequiredScreen/index.tsx
nklomp: Nit going to approve this change

nklomp: We went from  something not open to any misinterpretation to something that is open to it. The fact you can use numbers in truthy calls to your advantage doesn't mean you should.  So don't please

src/screens/SSICredentialsRequiredScreen/index.tsx
sanderPostma: Fixed, no idea where IJ got that the first time, just did ctrl-space

src/screens/SSICredentialsRequiredScreen/index.tsx
sanderPostma: @nklomp 
I changed it to a while ago
```
if (matchedVC?.length > 0) {
```
Isn't that ok?

nklomp: Just leave it as it was. That expression was done like that very much on purpose. You are now still relying on type coallescence in your bigger than expression. Basically resulting in undefined > 0. Which is false and thus correct, but still less readable then the original. Clarifty is always important and I simply don't like that we would go from something very clear to something less clear. Makes no sense at all to me

nklomp (approved):

PR #177, opened 2024-01-16
feature/WAL-686

BtencateSphereon:
PR requires updated ssi-sdk packages


src/screens/SSIContactAddScreen/index.tsx
```diff
@@ -110,7 +94,7 @@ class SSIContactAddScreen extends PureComponent<IProps, IState> {
       filter: [
         {
           contact: {
-            displayName: name,
+            legalName: name,
```
nklomp, line 97: Is this correct?

BtencateSphereon: displayName is not unique and currently we only support organizations

nklomp: Okay suggest to add a comment about that then

nklomp (commented):
See comment remark. Rest LGTM

src/screens/SSIContactAddScreen/index.tsx
BtencateSphereon: added a comment

nklomp (approved):
LGTM

PR #178, opened 2024-02-02
feature/VDX-318

BtencateSphereon:
requires a new release from the SSI-SDK


src/services/contactService.ts
```diff
@@ -1,53 +1,153 @@
...
+import {agentContext, cmAddContact, cmAddIdentity, cmGetContacts, cmRemoveContact, cmUpdateContact} from '../agent';
 import {IAddIdentityArgs, ICreateContactArgs, IGetContactsArgs, IUpdateContactArgs} from '../types';
 
-const debug: Debugger = Debug(`${APP_ID}:contactService`);
+const logger: EventLogger = new EventLoggerBuilder()
+  .withContext(agentContext)
+  .withLogLevel(LogLevel.INFO)
+  .withSystem(System.CONTACT)
+  .withSubSystem(SubSystem.CONTACT_MANAGER)
+  .withInitiatorType(InitiatorType.SYSTEM)
```
nklomp, line 12: You could change this to User, given most below methods are user initiated

src/services/contactService.ts
```diff
@@ -1,53 +1,153 @@
...
+  .build();
 
 export const getContacts = async (args?: IGetContactsArgs): Promise<Array<Party>> => {
-  debug(`getContacts(${JSON.stringify(args)})...`);
+  await logger.logEvent({
```
nklomp, line 16: This is an interesting one. In some cases this is user initiated. Then we really need to have these as audit events and at the INFO level. In a lot of cases this could also be system initiated, simply to get available contacts for a UI for instance. In that case it would be like it is currently (TRACE and system initiated)

src/services/contactService.ts
```diff
@@ -1,53 +1,153 @@
...
     })
     .catch((error: Error) => Promise.reject(Error(`Unable to create contact. Error: ${error}`)));
 };
 
 export const updateContact = async (args: IUpdateContactArgs): Promise<Party> => {
-  debug(`updateContact(${JSON.stringify(args)})...`);
+  await logger.logEvent({
+    type: LoggingEventType.AUDIT,
+    data: {
+      level: LogLevel.TRACE,
```
nklomp, line 66: This is not trace level

src/services/contactService.ts
```diff
@@ -1,53 +1,153 @@
...
+    },
+  });
+
   return cmAddContact(args)
     .then((contact: Party) => {
-      debug(`createContact(${JSON.stringify(args)}) succeeded`);
+      logger.logEvent({
+        type: LoggingEventType.AUDIT,
+        data: {
+          description: 'contact created',
```
nklomp, line 48: Wise to include the display name/alias in these calls

src/services/contactService.ts
```diff
@@ -1,53 +1,153 @@
...
     })
     .catch((error: Error) => Promise.reject(Error(`Unable to remove contact. Error: ${error}`)));
 };
 
 export const addIdentity = async (args: IAddIdentityArgs): Promise<Identity> => {
-  debug(`addIdentity(${JSON.stringify(args)})...`);
+  await logger.logEvent({
+    type: LoggingEventType.AUDIT,
+    data: {
+      level: LogLevel.TRACE,
```
nklomp, line 130: Not trace level; This one will also have a similar issue as the getContacts. Can be user initiated in the future, but also system initiated

src/services/credentialService.ts
```diff
@@ -15,30 +27,152 @@ import {
...
+  .withSystem(System.CREDENTIALS)
+  .withSubSystem(SubSystem.VC_PERSISTENCE)
+  .withInitiatorType(InitiatorType.SYSTEM)
+  .build();
+
 export const getVerifiableCredentialsFromStorage = async (): Promise<Array<UniqueVerifiableCredential>> => {
+  await logger.logEvent({
+    type: LoggingEventType.AUDIT,
+    data: {
+      level: LogLevel.TRACE,
```
nklomp, line 44: Same issue as getContacts

src/services/credentialService.ts
```diff
@@ -15,30 +27,152 @@ import {
...
+
   return dataStoreORMGetVerifiableCredentials();
 };
 
 export const storeVerifiableCredential = async (args: IStoreVerifiableCredentialArgs): Promise<string> => {
-  return dataStoreSaveVerifiableCredential({verifiableCredential: args.vc});
+  await logger.logEvent({
+    type: LoggingEventType.AUDIT,
+    data: {
+      level: LogLevel.TRACE,
```
nklomp, line 58: Not trace

src/services/credentialService.ts
```diff
@@ -15,30 +27,152 @@ import {
...
+
   return dataStoreGetVerifiableCredential({hash: args.hash});
 };
 
 export const deleteVerifiableCredential = async (args: IDeleteVerifiableCredentialArgs): Promise<boolean> => {
-  return removeCredentialBranding({filter: [{vcHash: args.hash}]}).then(() => dataStoreDeleteVerifiableCredential({hash: args.hash}));
+  await logger.logEvent({
+    type: LoggingEventType.AUDIT,
+    data: {
+      level: LogLevel.TRACE,
```
nklomp, line 105: Not trace

src/services/credentialService.ts
```diff
@@ -15,30 +27,152 @@ import {
...
+    })
+    .catch((error: Error) => Promise.reject(Error(`Unable to removed verifiable credential. Error: ${error}`)));
 };
 
 export const createVerifiableCredential = async (args: ICreateVerifiableCredentialArgs): Promise<VerifiableCredential> => {
-  return issueVerifiableCredential(args);
+  await logger.logEvent({
+    type: LoggingEventType.AUDIT,
+    data: {
+      level: LogLevel.TRACE,
```
nklomp, line 137: Not trace

nklomp (changes requested):
Some remarks.

Certain methods can be user initiated, like getCredentials, getContacts etc, but in a lot of cases they can also be system initiated. Then there is also a distinction at the level of logging. For system it would be trace, for user it would be info.  Probably wise to add a option to the call to make this distinction. We should make a ticket for that, as it makes little sense to resolve that problem now, given we still need to make sure we can get the actual user/identity with the logging call from a context anyway. As soon as we have that in place, we might be able to tell automatically.

src/services/credentialService.ts
BtencateSphereon: removed

src/services/credentialService.ts
BtencateSphereon: removed

src/services/credentialService.ts
BtencateSphereon: we do not have a way to see the difference

src/services/contactService.ts
BtencateSphereon: removed trace. and we do not have a way to see the difference

src/services/contactService.ts
BtencateSphereon: added the display name

src/services/contactService.ts
BtencateSphereon: removed

src/services/contactService.ts
BtencateSphereon: changed to USER

src/services/contactService.ts
nklomp: The simple approach would be to add an additional audit trial argument to the method, but lets revisit when we actually have authz in place

nklomp (approved):

PR #179, opened 2024-02-08
ibCredentialLocaleBrandingFrom error handling

sanderPostma:


src/providers/credential/OpenId4VcIssuanceProvider.ts
```diff
@@ -324,16 +324,22 @@ class OpenId4VcIssuanceProvider {
...
-            (metadata.display ?? []).map(
-              async (display: CredentialsSupportedDisplay): Promise<IBasicCredentialLocaleBranding> =>
-                await ibCredentialLocaleBrandingFrom({localeBranding: await credentialLocaleBrandingFrom(display)}),
-            ),
+          const optionalLocaleBranding: Array<IBasicCredentialLocaleBranding | undefined> = await Promise.all(
+            (metadata.display ?? []).map(async (display: CredentialsSupportedDisplay): Promise<IBasicCredentialLocaleBranding | undefined> => {
+              try {
+                return await ibCredentialLocaleBrandingFrom({localeBranding: await credentialLocaleBrandingFrom(display)});
+              } catch (e) {
+                console.error(`Could not fetch branding. Branding data: ${JSON.stringify(display)}`);
```
BtencateSphereon, line 336: i am not sure if this is the way we should handle this. Also what situation did you have where the branding was not present?

BtencateSphereon (changes requested):

src/providers/credential/OpenId4VcIssuanceProvider.ts
nklomp: Images could be unavailable because of a network issue for instance. I rather have us accept the credential without branding then having an error because of branding issues.

If I remember correctly it was a case where the image could not be found, right @sander

BtencateSphereon: should we then not fix that in the plugin instead of the wallet?

PR #185, opened 2024-04-24
feature/CWALL-166

BtencateSphereon:


robmuth:
Hey, it looks like there might be a little mistake in the package.json? https://github.com/Sphereon-Opensource/mobile-wallet/blob/86037fa66a9d827a7044f926c6be740c38e784c1/package.json#L63

nklomp:
Thanks for looking at it. This will be resolved when actually merging. It is the result of a story spanning multiple projects and libraries. We can progress with the review and then the different PRs will be merged and released in correct order

nklomp (approved):
I would have expected some more code to be gone, given a lot has moved to the SDK

PR #189, opened 2024-06-19
integrated credential-branding dep into mobile-wallet

sksadjad:
integrated credential-mapper dep into mobile-wallet


src/services/brandingService.ts
```diff
@@ -19,31 +17,3 @@ export const removeCredentialBranding = async (args: IRemoveCredentialBrandingAr
   debug(`removeCredentialBranding(${JSON.stringify(args)})...`);
   return ibRemoveCredentialBranding(args);
 };
-
-export const selectAppLocaleBranding = async (
```
nklomp, line 23: Did this get moved to ui-components or is it just removed?

sksadjad: yeah this is moved to the `credential-branding` module

nklomp (approved):

PR #191, opened 2024-07-02
Feature/sprind 18

sksadjad:
added saving and fetching issuer branding


src/components/views/SSIContactViewItem/index.tsx
```diff
@@ -9,24 +9,29 @@ import {
   SSIContactViewItemContactUriCaptionStyled as ContactUriCaption,
   SSIContactViewItemContainerStyled as Container,
   SSIContactViewItemLogoContainerStyled as LogoContainer,
-  SSIContactViewItemPlaceholderLogoStyled as PlaceholderLogo,
```
nklomp, line 12: Why is this removed? Not every issuer will have a logo

src/components/views/SSIContactViewItem/index.tsx
```diff
@@ -9,24 +9,29 @@ import {
...
-  const {name, uri, roles} = props;
-
+  const {name, uri, roles, logo} = props;
+  debug(`logo: ${logo ? JSON.stringify(logo) : ''}`);
   return (
     <Container>
       <StatusContainer />
       <LogoContainer>
-        <PlaceholderLogo />
+        <Logo logo={logo} />
```
nklomp, line 34: I would expect the placeholder logo to still be here in case no logo is available for the issuer

src/types/store/contact.types.ts
```diff
@@ -1,6 +1,9 @@
-import {Party} from '@sphereon/ssi-sdk.data-store';
+import {Party as StoreParty} from '@sphereon/ssi-sdk.data-store';
+import {ILocaleBranding} from '@sphereon/ssi-sdk.data-store';
 
 export interface IContactState {
   loading: boolean;
   contacts: Array<Party>;
 }
+
+export type Party = StoreParty & {branding?: ILocaleBranding};
```
nklomp: Why isn't this done in the SDK?

nklomp (changes requested):
See remarks

src/components/views/SSIContactViewItem/index.tsx
sksadjad: the SSILogo component which we're using has been doing the replacement by default, so there's no need to do it here

src/types/store/contact.types.ts
sksadjad: we want to manipulate types from ssi-sdk for usage in the wallet. not sure if we want that in the sdk level.

nklomp: Well you are simply combining 2 types from the same sdk module, and then using that accross multiple places in the wallet replacing a previous type from the SDK. As we have many places where the branding related to a contact makes sense en given they type combines 2 types from the same nodule I would expect that module to export the type

sksadjad: that makes sense to have it there. but do you also we can't support the db connection for it (because they belong to different modules and we don't necessarily use them together)

nklomp: Why couldn't that work?

There are many places accros modules where we call the other module via the context. Issuer branding depends on the contact plugin and store module. I am totally capable of calling contact module plugin methods from the branding module. Having said that I don't want us find out if we can move the logic to the SDK at this point, as we simply do not have the time at present. So suggest to only move the type to the SDK for now

nklomp: ping

sksadjad: yeah reverted this. now we're using the type from the ssi-sdk (which has the optional branding)

PR #192, opened 2024-07-05
Feature/sprind 18 fix reloading the credentials

sksadjad:


nklomp:
AFAIK we send events to referesh the list

PR #200, opened 2024-08-06
feature/SPRIND-31_musap-kms

sanderPostma:
Integrate MUSAP KMS into mobile wallet


android/app/build.gradle
```diff
@@ -153,11 +153,23 @@ android {
     }
 }
 
+// Access to MUSAP Maven artifacts
+repositories {
+    maven {
```
nklomp, line 158: Why is this being solved in here instead of in the RN module itself?

android/app/src/main/java/com/sphereon/ssi/wallet/MainApplication.kt
```diff
@@ -1,28 +1,32 @@
...
     override val reactNativeHost: ReactNativeHost = ReactNativeHostWrapper(
             this,
             object : DefaultReactNativeHost(this) {
-                override fun getPackages(): ArrayList<ReactPackage> {
-                    // Packages that cannot be autolinked yet can be added manually here, for example:
-                    // packages.add(new MyReactNativePackage());
-                    return PackageList(this).packages
+                override fun getPackages(): kotlin.collections.List<ReactPackage> {
+                    val packages = PackageList(this).packages.toMutableList()
+                    try {
```
nklomp, line 24: Same as above. The wallet should not be handling this. The RN package should

nklomp (changes requested):
Stopped review.

Why is the wallet doing all the changes that the RN package should be doing itself?

android/app/src/main/java/com/sphereon/ssi/wallet/MainApplication.kt
sanderPostma: In the React Native app I did not have to do this, for the expo app I do for reasons I don't know atm. So the standalone Rn app is auto-linking, our expo app is not.

android/app/build.gradle
sanderPostma: I've tried different ways, adding file repositories in the module's build.gradle doesn't seem to work. 
Considering the time we've already spent on this, I focused on getting it to work for our wallet for now.

android/app/src/main/java/com/sphereon/ssi/wallet/MainApplication.kt
sanderPostma: I can try to create an expo wrapper on the mac (since I can't create an expo lib on Windows,) but that will eat up more time.

nklomp: Yes but then at least create confluence page with things that need to be changed and create tickets linking to it. The credential store and Musap KMS are things we need long term, not PoC

android/app/build.gradle
nklomp: Yes but then at least create confluence page with things that need to be changed and create tickets linking to it. The credential store and Musap KMS are things we need long term, not PoC

sanderPostma: Created VDX-337 for now

android/app/src/main/java/com/sphereon/ssi/wallet/MainApplication.kt
sanderPostma: VDX-337

PR #204, opened 2024-08-06
Add new onboarding machine

### LINKED ISSUE #203: Move `SSIOnboardingTopBar` related files according to the project file structure

### LINKED ISSUE #205: Grab onboarding copy from i18n json files

### LINKED ISSUE #207: Create Stepper component

### LINKED ISSUE #208: Add "Show Progress" onboarding screen

odididi:
- Updated lockfile after fresh `yarn` on `funke`
- Extended `.prettierignore` directory list
- Added new onboarding machine with placeholder screens
- Removed references of the old onboarding machine
  - Removed tests
  - Removed guards related to field validations (Will be added in later commits)
  - Removed `setupWallet` service (Will be re-added when step 2 is fully implemented)
- Created `Stepper` component
- Added the `Show Progress` screen using the new `Stepper` component

Closes #203
Closes #205
Closes #207
Closes #208


src/machines/onboardingMachine.tsx
```diff
@@ -65,146 +117,22 @@ const createOnboardingMachine = (opts?: CreateOnboardingMachineOpts) => {
...
-        entry: assign({
-          pinCode: '',
-          personalData: undefined,
-          credentialData: undefined,
-          privacyPolicyAccepted: false,
-          termsConditionsAccepted: false,
-        }),
-      },
-    },
+    initial: OnboardingMachineStateType.showIntro,
```
BtencateSphereon, line 144: what happened to all the definitions for the events, guards and services? and would this pr also include the final step where we create the wallet and finish onboarding. also missing an error state, which we had some issues with in the past. but i am fine leaving that out for now and going for the happy path

src/components/bars/OnboardingHeader/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+      }}>
+      <HeaderRow>
+        <BackIconContainer style={{flex: 1}}>
+          <BackIcon
+            style={{marginTop: 0}}
+            icon={ButtonIconsEnum.BACK}
+            onPress={() => new Promise(() => onboardingInstance.send(OnboardingMachineEvents.PREVIOUS))}
+          />
+        </BackIconContainer>
+        {Boolean(title) && <SSITextH3LightStyled>{title as string}</SSITextH3LightStyled>}
```
BtencateSphereon: you don't have to use this Boolean.  if title is undefined it is falsy

src/screens/Onboarding/ShowProgressScreen/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+  return (
+    <Container>
+      <SSITextH0LightStyled>{translate(titleKey)}</SSITextH0LightStyled>
+      {descriptionKey && <SSITextH2LightStyled style={{opacity: 0.8, marginTop: 8}}>{translate(descriptionKey)}</SSITextH2LightStyled>}
+      <View style={{marginTop: 32, marginBottom: 'auto'}}>
+        <Stepper activeStep={currentStep - 1} content={stepperContent} ringColor={backgroundColors.primaryDark} />
+      </View>
+      <PrimaryButton
+        style={{height: 42, width: '100%'}}
+        caption="Next"
```
BtencateSphereon: would add localization here, we should have it already

BtencateSphereon (changes requested):
ok since it is for a POC I think it is fine as a base to work on. I left some small comments. The important one is about the state machine missing some states, we will need soon, and it seems this PR is about the new onboarding machine.

src/machines/onboardingMachine.tsx
```diff
@@ -1,47 +1,99 @@
...
+  enableBiometrics: {
+    on: {
+      NEXT: OnboardingMachineStateType.acceptTermsAndPrivacy,
+      PREVIOUS: OnboardingMachineStateType.verifyPinCode,
+    },
+  },
+  acceptTermsAndPrivacy: {
+    on: {
+      READ_TERMS: OnboardingMachineStateType.readTermsAndPrivacy,
+      PREVIOUS: OnboardingMachineStateType.enableBiometrics,
```
BtencateSphereon, line 95: does it make sense here to go back to the enableBiometrics screen? if the user has already enabled them, then we do not have a way of disabling them? and pressing maybe later results in just having them enabled?

src/machines/onboardingMachine.tsx
```diff
@@ -1,47 +1,99 @@
...
+  },
+  enableBiometrics: {
+    on: {
+      NEXT: OnboardingMachineStateType.acceptTermsAndPrivacy,
+      PREVIOUS: OnboardingMachineStateType.verifyPinCode,
+    },
+  },
+  acceptTermsAndPrivacy: {
+    on: {
+      READ_TERMS: OnboardingMachineStateType.readTermsAndPrivacy,
```
BtencateSphereon: i guess this is for showing the terms itself? should we not have one for the privacy as well?

src/machines/onboardingMachine.tsx
```diff
@@ -1,47 +1,99 @@
...
+  verifyPinCode: {
+    on: {
+      NEXT: OnboardingMachineStateType.enableBiometrics,
+      PREVIOUS: OnboardingMachineStateType.enterPinCode,
+    },
+  },
+  enableBiometrics: {
+    on: {
+      NEXT: OnboardingMachineStateType.acceptTermsAndPrivacy,
+      PREVIOUS: OnboardingMachineStateType.verifyPinCode,
```
BtencateSphereon: hmmm i don't think this should go to verifyPinCode, but to enterPinCode.

src/machines/onboardingMachine.tsx
BtencateSphereon: also it seems to miss step 3 and onwards. i am not sure if want to layout the whole machine or just a part of it

odididi: As i mentioned in the [description](https://github.com/Sphereon-Opensource/mobile-wallet/pull/204#issue-2451020063), these were removed as they are not used or should be defined in the current state of the onboarding machine. They should be re-introduced when we implement the relevant screens.

odididi: Merveille will be adding the new states as he implements step 3 (import personal data)

src/components/bars/OnboardingHeader/index.tsx
odididi: I know it works without here. It is a general best practice in any conditional rendering as a habit to avoid leaked rendering. Will remove it though, since we don't have any specific lint rules regarding it.

src/screens/Onboarding/ShowProgressScreen/index.tsx
odididi: 👍🏻

src/machines/onboardingMachine.tsx
BtencateSphereon: hmmm, i guess if you press maybe later, and you change your mind, you can enable them. so there are multiple flows with different outcomes we need to check i guess

src/machines/onboardingMachine.tsx
odididi: You are right. I didnt notice the two separate screens for them, and assumed they were combined. I will add the extra states

src/machines/onboardingMachine.tsx
odididi: Indeed the back functionality should be discussed it seems. For now i just defaulted it to the previous state.

src/machines/onboardingMachine.tsx
odididi: ditto

src/machines/onboardingMachine.tsx
BtencateSphereon: but you are using events, and it would be nice if we turn the conditions into actual guards like we do within the current machines. i am also fine to say ok, we can do that later, but then we need to make sure that happens

src/screens/Onboarding/ShowProgressScreen/index.tsx
odididi: These will be updated eventually per the designs. For now ill change them all to the specified translation label

src/machines/onboardingMachine.tsx
odididi: 👍🏻

odididi: added two guards

src/components/bars/OnboardingHeader/index.tsx
odididi: removed Boolean

src/screens/Onboarding/ShowProgressScreen/index.tsx
odididi: Updated

src/machines/onboardingMachine.tsx
odididi: Fixed. Query params pass down the state, so a user can see the relevant document when visiting the "read" page

BtencateSphereon (approved):
LGTM

PR #209, opened 2024-08-08
Feat/SPRIND-27

Brummos:


src/providers/authentication/eIDProvider.ts
```diff
@@ -0,0 +1,116 @@
+import {Dispatch, SetStateAction} from 'react';
+import {addMessageListener, AusweisAuthFlow, AusweisSdkMessage, sendCommand} from '@animo-id/expo-ausweis-sdk';
+import {PARMode} from '@sphereon/oid4vci-common';
+import {OpenID4VCIClient} from '@sphereon/oid4vci-client';
+import {EIDFlowState, EIDGetAuthorizationCodeArgs, EIDHandleErrorArgs, EIDInitializeArgs, EIDProviderArgs} from '../../types';
+
+class EIDProvider {
```
nklomp: Suggest to call this something like a PIDServiceGermany. It is all about PIDs

src/providers/authentication/eIDProvider.ts
```diff
@@ -0,0 +1,116 @@
...
+      retrieveServerMetadata: true,
+      clientId: 'bc11dd24-cbe9-4f13-890b-967e5f900222',
+      // This is a separate call, so we don't fetch it here, however it may be easier to just construct it here?
+      createAuthorizationRequestURL: false,
+    });
+
+    const authorizationRequestUrl = await client.createAuthorizationRequestUrl({
+      authorizationRequest: {
+        redirectUri: 'https://sphereon.com/redirect',
+        scope: 'pid', // TODO we need to determine the scopes using the metadata
```
nklomp: No, since this Provider is tied to the German PID Issuer, we actually better have a fixed scope

src/providers/authentication/eIDProvider.ts
```diff
@@ -0,0 +1,116 @@
...
+        //authorizationDetails: authDetails,
+        parMode: PARMode.AUTO,
```
nklomp: Required

nklomp (changes requested):
Some small remarks

src/providers/authentication/eIDProvider.ts
BtencateSphereon: changed to PIDServiceGermany

src/providers/authentication/eIDProvider.ts
BtencateSphereon: removed the comment

src/providers/authentication/eIDProvider.ts
BtencateSphereon: changed to REQUIRE

nklomp (approved):
LGTM

PR #210, opened 2024-08-12
feature/SPRIND-35

BtencateSphereon:


src/screens/CredentialCatalogScreen/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+      title: 'Girocard',
+      issuer: 'EURO Kartensysteme GmbH',
+      branding: {
+        background: {
+          image: {
+            uri: 'https://i.ibb.co/NngL9SD/giro-Card-Background.png',
+          },
+        },
+        logo: {
+          uri: 'https://i.ibb.co/P928gkN/giro-Card-Logo.png',
```
nklomp: Could we download these and provide as assets. Don't want to stand in a venue with choppy internet connection and the screen loading all these images slowly. It is a mock/demo for the most part anyway

nklomp (changes requested):
See remark about demo mini-cards being fetched online

nklomp:
Also some conflicts must be resolved

nklomp (approved):

PR #251, opened 2024-08-27
Feat/onboarding touch ups

luukschipperheyn:
- fix ios bottom padding - also with keyboard in view
- make textfield underline white instead of gradient
- on invalid. pin entry, smaller screens scroll down to feedback msg


luukschipperheyn:
Messed something up while rebasing

PR #256, opened 2024-08-28
Fix/empty pin entry on navigate

luukschipperheyn:
- reset pincode textinput value in EnterPincodeScreen to empty string upon navigating forward
- focus pincode textinput upon navigating back to EnterPincodeScreen


luukschipperheyn:
hold on. found a bug

PR #278, opened 2024-09-27
fix/c2 flow fixes

BtencateSphereon:


nklomp (approved):

src/services/pexService.ts
```diff
@@ -123,7 +123,7 @@ export const getMatchingPidCredentials = async ({
   const credentials = pidCredentials.map(pidCredential => {
     const dc: NonPersistedDigitalCredential = nonPersistedDigitalCredentialEntityFromAddArgs({
       rawDocument: pidCredential.rawCredential,
-      kmsKeyRef: 'fixme', // FIXME
+      kmsKeyRef: 'https://demo.pid-issuer.bundesdruckerei.de/c2/.well-known/openid-credential-issuer',
```
nklomp: How does this work when using the regular C flow?

nklomp (changes requested):
Not sure the C flow still works mainly when sharing with a RP. If you ask me the static kmsKeyRef will cause issues

src/services/pexService.ts
BtencateSphereon: not ignored, but i am waiting on a new ssi-sdk version to test everything

BtencateSphereon: Seems to be working for in all flows. but i made it a condition to only set the kmsKeyRef for the c2 flow. would that be ok for now?

PR #281, opened 2024-10-01
feature/sprind 72 biometrcis on login

merveillevaneck:
- added fixed for locking handler with biometrics, added bio login
- added useEnableBiometrics for easy toggling, added setBiometrics thunk


BtencateSphereon (changes requested):
biometrics is not working / showing up when you reboot the app

PR #287, opened 2024-10-25
feature/SPHEREON-1183_sd-jwt-conformance

sanderPostma:
support disclosure of entire sd-jwt groups which have individual frame elements


zoemaas (approved):
LGTM

PR #290, opened 2024-10-29
feature/MWALL-709_funke-only-de

sanderPostma:


src/types/countries/index.ts
```diff
@@ -0,0 +1,69 @@
+import {CountryOption} from '../component';
```
BtencateSphereon: this file should be moved to utils

BtencateSphereon (changes requested):
problem with displaying country names, sorting country names. and 1 small remark

src/types/countries/index.ts
sanderPostma: fixed

PR #292, opened 2024-10-31
feat/sprind 96 contact updates

merveillevaneck:
- added new contact screens, contact header, and ssi-sdk updates for branding
- various contact details updates

this includes an update to ssi-sdk deps to get new typings from ssi-sdk.data-store for the contact branding. 

prupose of the pr is to sync up with various team members and specifically merge a the ContactInformationView component so that it can be used in other places as well.

This pr includes the above component but also includes skeletons for the contact identities screen as well as contact activities. these screen are not complete, but need to be merged as warp has some components they want to add to these screens that have already been made by odi.


src/screens/SSIContactsOverviewScreen/index.tsx
```diff
@@ -74,7 +75,7 @@ class SSIContactsOverviewScreen extends PureComponent<IProps, IState> {
     };
     const style = {
       ...backgroundStyle,
-      ...(itemInfo.index === contacts.length - 1 && itemInfo.index % 2 !== 0 && {borderBottomWidth: 1, borderBottomColor: borderColors.dark}),
+      // ...(itemInfo.index === contacts.length - 1 && itemInfo.index % 2 !== 0 && {borderBottomWidth: 1, borderBottomColor: borderColors.dark}),
```
BtencateSphereon: why has this been disabled?

BtencateSphereon (changes requested):

src/screens/SSIContactsOverviewScreen/index.tsx
merveillevaneck: i am unsure whether the new contact designs require a bottom border on the last item of the list. but i have added this bottom border back in now. i dont see that it would be an issue to leave it in.

PR #299, opened 2024-11-11
feature/holder qr presentation

merveillevaneck:
- implemented QR code presentation screen skeleton and nav button in header
- added qr code generator dep, added to agent, created share stack for holder
- added mock variable for temp display on the QR Presentation screen

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@sphereon/ssi-sdk.issuance-branding": "0.30.2-next.191",
     "@sphereon/ssi-sdk.mdl-mdoc": "0.30.2-next.191",
     "@sphereon/ssi-sdk.oid4vci-holder": "0.30.2-next.191",
-    "@sphereon/ssi-sdk.qr-code-generator": "^0.30.1",
+    "@sphereon/ssi-sdk.qr-code-generator": "0.30.2-next.191",
     "@sphereon/ssi-sdk.sd-jwt": "0.30.2-next.191",
     "@sphereon/ssi-sdk.siopv2-oid4vp-op-auth": "0.30.2-next.191",
     "@sphereon/ssi-sdk.vc-handler-ld-local": "0.30.2-next.191",
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "@sphereon/ssi-sdk.issuance-branding": "0.30.2-next.191",
     "@sphereon/ssi-sdk.mdl-mdoc": "0.30.2-next.191",
     "@sphereon/ssi-sdk.oid4vci-holder": "0.30.2-next.191",
+    "@sphereon/ssi-sdk.qr-code-generator": "^0.30.1",
     "@sphereon/ssi-sdk.sd-jwt": "0.30.2-next.191",
     "@sphereon/ssi-sdk.siopv2-oid4vp-op-auth": "0.30.2-next.191",
     "@sphereon/ssi-sdk.vc-handler-ld-local": "0.30.2-next.191",
--- a/src/components/bars/SSIHeaderBar/index.tsx
+++ b/src/components/bars/SSIHeaderBar/index.tsx
@@ -69,7 +69,7 @@ const SSIHeaderBar: FC<HeaderBarProps> = (props: HeaderBarProps): JSX.Element =>
   };
 
   const onNavigateQrPresentation = () => {
-    navigation.navigate(ScreenRoutesEnum.QR_PRESENTATION);
+    navigation.navigate(MainRoutesEnum.SHARE);
   };
 
   const onMore = async (): Promise<void> => {
--- a/src/components/bars/SSIHeaderBar/index.tsx
+++ b/src/components/bars/SSIHeaderBar/index.tsx
@@ -121,13 +121,12 @@ const SSIHeaderBar: FC<HeaderBarProps> = (props: HeaderBarProps): JSX.Element =>
               <SSIDropDownList
                 buttons={[
                   {
-                    // caption: translate('profile_logout_action_caption'),
-                    caption: 'Settings',
+                    caption: translate('settings_dropdown_item_text'),
                     onPress: onNavigateProfile,
                     icon: HeaderMenuIconsEnum.SETTINGS,
                   },
                   {
-                    caption: 'Present QR code',
+                    caption: translate('present_qr_code_dropdown_item_text'),
                     onPress: onNavigateQrPresentation,
                     icon: HeaderMenuIconsEnum.QR,
                   },
--- a/src/components/bars/SSIHeaderBar/index.tsx
+++ b/src/components/bars/SSIHeaderBar/index.tsx
@@ -129,7 +129,7 @@ const SSIHeaderBar: FC<HeaderBarProps> = (props: HeaderBarProps): JSX.Element =>
                   {
                     caption: 'Present QR code',
                     onPress: onNavigateQrPresentation,
-                    // icon: HeaderMenuIconsEnum.
+                    icon: HeaderMenuIconsEnum.QR,
                   },
                 ]}
               />
--- a/src/components/bars/SSIHeaderBar/index.tsx
+++ b/src/components/bars/SSIHeaderBar/index.tsx
@@ -19,7 +19,7 @@ import {
   SSIRightColumnRightAlignedContainerStyled as RightColumn,
   SSIFlexDirectionRowViewStyled as Row,
 } from '../../../styles/components';
-import {ButtonIconsEnum, HeaderMenuIconsEnum, IHeaderMenuButton, IUser, MainRoutesEnum, StackParamList} from '../../../types';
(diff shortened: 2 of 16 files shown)


src/components/bars/SSIHeaderBar/index.tsx
```diff
@@ -122,6 +126,11 @@ const SSIHeaderBar: FC<HeaderBarProps> = (props: HeaderBarProps): JSX.Element =>
                     onPress: onNavigateProfile,
                     icon: HeaderMenuIconsEnum.SETTINGS,
                   },
+                  {
+                    caption: 'Present QR code',
```
BtencateSphereon: needs localization

src/components/dropDownLists/SSIDropDownListItem/index.tsx
```diff
@@ -45,6 +46,8 @@ const SSIDropDownListItem: FC<IProps> = (props: IProps): JSX.Element => {
         return <SSIDownloadIcon color={fontColor} />;
       case HeaderMenuIconsEnum.SETTINGS:
         return <SettingsIcon color={fontColor} />;
+      case HeaderMenuIconsEnum.QR:
+        return <SSIQRIcon primaryColor={fontColor.toString()} secondaryColor={fontColor.toString()} />;
```
BtencateSphereon: why is there a toString used if other icons do not need this? please keep consistency

src/navigation/navigation.tsx
```diff
@@ -943,6 +954,22 @@ export const FunkeC2ShareStackScreenWithContext = (props: any): JSX.Element => (
...
 
+const ShareStack = (): JSX.Element => {
+  return (
+    <ShareBaseStack.Navigator screenOptions={{animation: 'none'}} initialRouteName={ScreenRoutesEnum.QR_PRESENTATION}>
+      <ShareBaseStack.Screen
+        name={ScreenRoutesEnum.QR_PRESENTATION}
+        component={QRPresentationScreen}
+        options={{
+          header: props => <SSIHeaderBar headerSubTitle="Show the QR code to the party of interest. So they can scan your QR code." {...props} />,
+          headerTitle: 'Present QR code',
```
BtencateSphereon: needs localization

src/navigation/navigation.tsx
```diff
@@ -943,6 +954,22 @@ export const FunkeC2ShareStackScreenWithContext = (props: any): JSX.Element => (
...
 );
 
+const ShareStack = (): JSX.Element => {
+  return (
+    <ShareBaseStack.Navigator screenOptions={{animation: 'none'}} initialRouteName={ScreenRoutesEnum.QR_PRESENTATION}>
+      <ShareBaseStack.Screen
+        name={ScreenRoutesEnum.QR_PRESENTATION}
+        component={QRPresentationScreen}
+        options={{
+          header: props => <SSIHeaderBar headerSubTitle="Show the QR code to the party of interest. So they can scan your QR code." {...props} />,
```
BtencateSphereon: needs localization

package.json
```diff
@@ -72,6 +72,7 @@
     "@sphereon/ssi-sdk.issuance-branding": "0.30.2-next.191",
     "@sphereon/ssi-sdk.mdl-mdoc": "0.30.2-next.191",
     "@sphereon/ssi-sdk.oid4vci-holder": "0.30.2-next.191",
+    "@sphereon/ssi-sdk.qr-code-generator": "^0.30.1",
```
BtencateSphereon: should have the same version as other ssi-sdk packages to avoid any issues

src/screens/QRPresentationScreen/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+        size: (3 * width) / 4,
+        title: 'Presentation',
+      },
+    };
+
+    setQrElement(await agentContext.agent.qrURIElement(uriElementArgs));
+  };
+
+  useEffect(() => {
+    delegate();
```
BtencateSphereon: return value is ignored, you can void this

src/screens/QRPresentationScreen/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type QRPresentationScreenProps = {};
+
+/* FIXME: Replace this with some result once back end for holder
+ * presentations is complete.
+ */
+const mockURI = 'some_uri_content';
+
+const QRPresentationScreen: FC<QRPresentationScreenProps> = (props: QRPresentationScreenProps) => {
+  const [qrElement, setQrElement] = useState<React.ReactElement | null>(null);
```
BtencateSphereon: please import React instead of referencing it

src/screens/QRPresentationScreen/index.tsx
```diff
@@ -0,0 +1,53 @@
+import {FC, useEffect, useState} from 'react';
+import {Dimensions} from 'react-native';
+import {SSIBasicContainerSecondaryStyled as SSIContainer} from '../../styles/components';
+import {QRContainer, QRPresentationViewContainer} from 'src/styles/components/screens/QRPresentationScreen';
+import {CreateElementArgs, QRType} from '@sphereon/ssi-sdk.qr-code-generator';
+import {agentContext} from '../../agent';
+
+const {width} = Dimensions.get('screen');
+
+type QRPresentationScreenProps = {};
```
BtencateSphereon: for now you can just not add any props, so this type can be removed. we can add it once we need it

src/navigation/navigation.tsx
```diff
@@ -943,6 +954,22 @@ export const FunkeC2ShareStackScreenWithContext = (props: any): JSX.Element => (
...
+          headerTitle: 'Present QR code',
+          headerShown: true,
```
BtencateSphereon: can be omitted, default is true

BtencateSphereon: https://reactnavigation.org/docs/stack-navigator/#headershown

src/navigation/navigation.tsx
```diff
@@ -184,6 +187,14 @@ const MainStackNavigator = (): JSX.Element => {
           </>
         )}
       />
+      <Stack.Screen
+        name={MainRoutesEnum.SHARE}
+        children={() => (
+          <>
```
BtencateSphereon: we do not need the empty tags, because there is only 1 child

BtencateSphereon (changes requested):

src/screens/QRPresentationScreen/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+      data: {
+        type: QRType.URI,
+        object: mockURI,
+        //FIXME: determine if the id field should be necessary on a URI
+        //element
+        id: '',
+      },
+      renderingProps: {
+        bgColor: 'white',
+        fgColor: '#352575',
```
BtencateSphereon: design shows the color black

BtencateSphereon (changes requested):

src/components/dropDownLists/SSIDropDownListItem/index.tsx
merveillevaneck: i have made an update to this by changing the prop types of the `SSIQRIcon` component to use `ColorValue` from react-native instead of `string`. Since i noticed this is what the other icon components are doing. So i have updated the usage of this to use `fontColor` directly.

I believe the reason why there are primary and secondary colors on this component already is because it supports gradient colors.

src/navigation/navigation.tsx
merveillevaneck: added localization here

src/navigation/navigation.tsx
merveillevaneck: added localization for this

package.json
merveillevaneck: ive update the version of the package to `0.30.2-next.191`

src/screens/QRPresentationScreen/index.tsx
merveillevaneck: added `void delegate()`

src/screens/QRPresentationScreen/index.tsx
merveillevaneck: I ended up directly importing `ReactElement`

src/screens/QRPresentationScreen/index.tsx
merveillevaneck: props have been removed

src/navigation/navigation.tsx
merveillevaneck: I have removed this `headerShown` option from the object

src/navigation/navigation.tsx
merveillevaneck: I have removed these tags

src/screens/QRPresentationScreen/index.tsx
merveillevaneck: I have changed this color to `'black'`

src/components/bars/SSIHeaderBar/index.tsx
merveillevaneck: added localization here. I also saw that there was no localization for the settings dropdown item. so I added a localization string for that one as well.

PR #304, opened 2024-11-12
feature/SPRIND-120

BtencateSphereon:


src/services/machines/siopV2MachineService.ts
```diff
@@ -189,3 +189,45 @@ export const sendResponse = async (
...
+
+  if (!authorizationRequestData?.uri) {
+    return Promise.reject(Error('Missing authorization request data  uri in context'));
+  }
+
+  const parsedUrl = new URL(url);
+  const params = new URLSearchParams(parsedUrl.search);
+  const openidFederation = params.get('openid_federation');
+
+  const entityIdentifier = openidFederation ?? authorizationRequestData.clientId;
```
sanderPostma: It's never going to clientId anymore but entityId. But you don't have that in the lib interface yet. 
And openid_federation is not a known query param and clientId is not valid for openid-federation requests.
I guess this ticket is impediment until SDK-41 is done.

src/services/machines/siopV2MachineService.ts
```diff
@@ -189,3 +189,45 @@ export const sendResponse = async (
...
+    return Promise.reject(Error('Missing request url in context'));
+  }
+
+  if (!authorizationRequestData?.uri) {
+    return Promise.reject(Error('Missing authorization request data  uri in context'));
+  }
+
+  const parsedUrl = new URL(url);
+  const params = new URLSearchParams(parsedUrl.search);
+  const openidFederation = params.get('openid_federation');
```
sanderPostma: openid_federation is not listed as a valid query parameter in the OpenID4VP specification.

sanderPostma (changes requested):

src/services/machines/siopV2MachineService.ts
BtencateSphereon: yes, i think i will remove this, as it seems we have a spec for this and not some extension that was made specifically for this. this is coming from the vci solution

nklomp (approved):
LGTM

PR #315, opened 2024-11-20
feature/SPRIND-116

zoemaas:


src/types/machines/siopV2/index.ts
```diff
@@ -36,6 +38,11 @@ export type SiopV2MachineContext = {
   contactAlias: string;
   selectedCredentials: Array<UniqueDigitalCredential>;
   trustedAnchors?: Array<TrustedAnchor>;
+  federation_entity?: any
```
sanderPostma, line 42: Can't federation_entity have a type here?

sanderPostma (changes requested):

package.json
```diff
@@ -59,7 +59,7 @@
     "@sphereon/ssi-sdk-ext.did-resolver-jwk": "0.26.1-next.11",
     "@sphereon/ssi-sdk-ext.did-resolver-key": "0.26.1-next.11",
     "@sphereon/ssi-sdk-ext.did-utils": "0.26.1-next.11",
-    "@sphereon/ssi-sdk-ext.identifier-resolution": "0.26.1-next.11",
+    "@sphereon/ssi-sdk-ext.identifier-resolution": "file:../SSI-SDK-crypto-extensions/packages/identifier-resolution",
```
sanderPostma: Do not forget to revert to a real version before merge

src/types/machines/siopV2/index.ts
zoemaas: Yes and no. Yes because we have an @sphereon/openid-federation-open-api and no because the library does not contain the openid federation types. Since duplication is not a good choice, I opened a bug ticket to fix this.

package.json
zoemaas: Updated it to use the package from the remote npm registry

sanderPostma (approved):

PR #342, opened 2024-12-03
Feature/oidf 69#2

jcmelati:


src/components/bars/OnboardingHeader/index.tsx
```diff
@@ -133,6 +134,9 @@ const OnboardingHeader: FC<HeaderBarProps> = ({title, stepConfig, onBack, header
             type: OnboardingMachineEvents.UPDATE_SECURITY_MODEL,
             model: securityModel,
           });
+          if(securityModel === PIDSecurityModel.REMOTE_HSM) {
```
sanderPostma: I think this is duplicate now, is done already in storagePersistPIDSecurityModel

sanderPostma (approved):
Just remove the dup code. I approved already because that's not critical and I won' need to come back here again

PR #347, opened 2024-12-04
fix: onboarding navigation back issue

jcmelati:


src/machines/onboardingMachine.tsx
```diff
@@ -348,7 +348,7 @@ const states: OnboardingStatesConfig = {
   },
   declinePIDCredentials: {
     on: {
-      PREVIOUS: OnboardingMachineStateType.reviewPIDCredentials,
```
BtencateSphereon: this needs to be reverted, as decline pid is a seperate screen. if a user changes it's mind the user now needs to start over with no change to the PID. the real decline action happens on the NEXT of this state

src/screens/Onboarding/CompleteOnboardingScreen/index.tsx
```diff
@@ -57,15 +57,6 @@ const CompleteOnboardingScreen = () => {
           </Svg>
         )}
       </View>
-      <ExitButtonContainer onPress={() => onboardingInstance.send(OnboardingMachineEvents.PREVIOUS)}>
```
BtencateSphereon, line 60: please cleanup ExitButtonContainer at the top of the file

BtencateSphereon: also cleanup the imports

BtencateSphereon, line 60: to make it more robust we need to add a BackHandler to this screen, if a native back event is fired we should just continue to the app. because currently this would break the flow

we already use it in the app on some places but we need to change the logic so that we fire the 
onboardingInstance.send(OnboardingMachineEvents.NEXT) on a back event

```
  useBackHandler((): boolean => {
    onboardingInstance.send(OnboardingMachineEvents.NEXT) 
    return true;
  });
```

BtencateSphereon (changes requested):

BtencateSphereon:
Also please make sure you apply the same changes to the getPIDCredentials machine and the FunkeC2 machine

src/screens/Onboarding/CompleteOnboardingScreen/index.tsx
BtencateSphereon: and as an additional question. do we need to update the machine to not have a PREVIOUS event in this state?

PR #365, opened 2025-01-27
feature/SPHEREON-1196

BtencateSphereon:


src/services/machines/onboardingMachineService.ts
```diff
@@ -121,20 +126,77 @@ export const storePIDCredentials = async (context: Pick<OnboardingMachineContext
...
       value: pinCode,
     }),
+    createSelfIssuedCredential(context)
+        .then((credential) =>
+           agent.ibAddCredentialBranding({
+            vcHash: credential.hash,
+            issuerCorrelationId: credential.issuerCorrelationId,
+            localeBranding: [SphereonWalletIdentityBranding]
+          })
+        ),
```
sanderPostma, line 143: what happens if createSelfIssuedCredential (or storagePersistPin)fails? Can the user go back and try again?
Not needed to do a .catch() block which returns a Promise.reject() ?

src/services/machines/onboardingMachineService.ts
```diff
@@ -121,20 +126,77 @@ export const storePIDCredentials = async (context: Pick<OnboardingMachineContext
...
+        ...(names.lastName && { lastName: names.lastName })
+      },
+    },
+    proofFormat: credentialData.proofFormat ?? 'jwt',
+    header: {
+      kid: key?.meta?.verificationMethod?.id,
+    },
+  });
+  return storeVerifiableCredential({
+    credentialRole: CredentialRole.HOLDER, // Here we are both ISSUER & HOLDER but has I think it to be HOLDER due to "oid4vp.filterCredentialsAgainstAllDefinitions(CredentialRole.HOLDER)"
```
sanderPostma: When are storing it, we are holder

src/types/machines/onboarding/index.ts
```diff
@@ -8,6 +8,15 @@ import {OnboardingStackParamsList} from '../../navigation';
 import {IUser} from '../../user';
 import {MappedCredential} from '../getPIDCredentialMachine';
 import {PIDSecurityModel} from '../../../services/storageService';
+import {CredentialPayload, ProofFormat} from '@veramo/core';
+import {SupportedDidMethodEnum} from '../../did';
+
+export type OnboardingCredentialData = {
+  didMethod: SupportedDidMethodEnum;
+  didOptions?: any;
```
sanderPostma, line 16: Is this one needed? Since IDIDOptions is deprecated.

src/services/machines/onboardingMachineService.ts
```diff
@@ -121,20 +126,77 @@ export const storePIDCredentials = async (context: Pick<OnboardingMachineContext
...
+  const {emailAddress, name, credentialData} = context;
+  const identifier: IIdentifier = await getOrCreatePrimaryIdentifier(
+      {
+        method: credentialData.didMethod,
+        createOpts: {options: credentialData.didOptions},
+      },
+      agentContext,
+  );
+
+  const names = parseFullName(name);
```
sanderPostma, line 165: Can this be undefined? Or fields within names.
(I cannot see in parseFullName here)

sanderPostma (changes requested):
No biggies, but things to double-check

src/services/machines/onboardingMachineService.ts
BtencateSphereon: no default is an empty string in the onboarding machine

src/types/machines/onboarding/index.ts
BtencateSphereon: i took it from whatever we were doing before, so for now i would say, address it later if needed

src/services/machines/onboardingMachineService.ts
BtencateSphereon: this would be handled by the onboarding machine. as this results in a error and thus the invoke onError gets triggered in the machine.

not saying this is being handled gracefully as i have not checked, but i would say this is out of scope of this ticket

to clarify, if any of the promises results in an error, the whole block immediately rejects. sure we can improve the unhappy flow. and make sure that if we retry that we do not have 2 vc's for example. but for now i think this is fine

BtencateSphereon: https://sphereon.atlassian.net/browse/MWALL-725

src/services/machines/onboardingMachineService.ts
BtencateSphereon: it was your own comment. removed as we are indeed the holder

sanderPostma (approved):
LGTM

PR #379, opened 2025-09-30
feature/SSISDK-41

BtencateSphereon:


src/components/views/CredentialSelectView/index.tsx
```diff
@@ -77,8 +76,16 @@ const CredentialSelectView = (props: CredentialSelectViewProps) => {
 
   const onPressCredential = async (credential: UniqueDigitalCredential): Promise<void> => {
     onSelect(credential);
-    setSelectedCredential(credential);
-    await loadCredentialContent(credential, dcqlQuery);
+    setSelectedCredential(prev => {
+      if (prev?.hash === credential.hash && credentials.length > 1) {
```
sanderPostma, line 80: Please add a little comment what is going on here. (If you don't know the context)

src/components/views/CredentialSelectView/index.tsx
```diff
@@ -77,8 +76,16 @@ const CredentialSelectView = (props: CredentialSelectViewProps) => {
...
+        return null;
+      }
+      return credential;
+    });
+
+    if (selectedCredential?.hash !== credential.hash) {
```
sanderPostma, line 87: Please add a little comment what is going on here. (If you don't know the context)

src/screens/CredentialOverviewShareScreen/index.tsx
```diff
@@ -1,73 +1,58 @@
...
     credentials.map((vc) => [convertToDcqlCredentials(vc), vc])
   )
 
   const queryResult = DcqlQuery.query(dcqlQuery, Array.from(dcqlCredentialsWithCredentials.keys()))
 
   const selectableCredentialsMap = new Map()
-  for (const [key, value] of Object.entries(queryResult.credential_matches)) {
-    if (!value.valid_credentials) {
-      continue
+  if (queryResult.credential_sets) {
```
sanderPostma, line 40:
```suggestion
/**
 * DCQL → UI mapping:
 * Populate `selectableCredentialsMap` from a DCQL `queryResult`.
 * - With `credential_sets`: map each set index to all local credentials that satisfy any option.
 * - Without sets: map each requirement key to its validated local credentials.
 * Result: per-requirement pick lists of eligible credentials for the UI.
 */
  if (queryResult.credential_sets) {
```

src/screens/CredentialOverviewShareScreen/index.tsx
```diff
@@ -1,73 +1,58 @@
...
+    queryResult.credential_sets.forEach((credentialSet, index) => {
+      if (!credentialSet.matching_options) {
+        return
+      }
+      const credentialSetMatches: Array<UniqueDigitalCredential> = [];
+      credentialSet.matching_options.forEach(options => {
+        options.flat().forEach(option => {
+          const matchedCredentials = queryResult.credential_matches[option].valid_credentials?.map(cred => credentials[cred.input_credential_index]) ?? []
```
sanderPostma: Not like 
queryResult.credential_matches[option]?.valid_credentials?
?

src/screens/CredentialOverviewShareScreen/index.tsx
```diff
@@ -1,73 +1,58 @@
...
-  //     selectedCredential.originalVerifiableCredential = result.verifiableCredential?.[i];
-  //     subsetCredentials.push(selectedCredential);
-  //   }
-  // }
-
-  return []
-};
+type GetCredentialSelectViewElement = {
+  index: number
+  itemId: string | number
```
sanderPostma, line 20: Is that safe here? Do we always handle this well also when the credential id would be '1234'? Or would we need a discriminator or something?

src/components/views/CredentialSelectView/index.tsx
```diff
@@ -31,8 +31,7 @@ import { DcqlQuery } from 'dcql';
 
 type CredentialSelectViewProps = {
   onSelect: (credential: UniqueDigitalCredential) => void;
-  credentials: UniqueDigitalCredential[];
-  //presentationDefinition: IPresentationDefinition;
+  credentials: Array<UniqueDigitalCredential>;
```
sanderPostma: ReadonlyArray<UniqueDigitalCredential> ?

src/screens/CredentialOverviewShareScreen/index.tsx
```diff
@@ -1,73 +1,58 @@
...
   const selectableCredentialsMap = new Map()
-  for (const [key, value] of Object.entries(queryResult.credential_matches)) {
-    if (!value.valid_credentials) {
-      continue
+  if (queryResult.credential_sets) {
+    queryResult.credential_sets.forEach((credentialSet, index) => {
+      if (!credentialSet.matching_options) {
+        return
+      }
+      const credentialSetMatches: Array<UniqueDigitalCredential> = [];
```
sanderPostma, line 45: Shouldn't this be a hash set or something? I mean can't we accidentally push the same credential twice when going through a larger rules set?

src/screens/CredentialOverviewShareScreen/index.tsx
sanderPostma, line 20: Can we safely handle all use cases using this union type? Also when someone names a credential id '1234'? Or do we need some kind of discriminator?

sanderPostma (changes requested):
Some questions & suggestion

src/screens/CredentialOverviewShareScreen/index.tsx
BtencateSphereon: this comes from the fact that credentials in a dcql query have an id (which is a string), and for credential_sets we only have the index of the array to work with

src/screens/CredentialOverviewShareScreen/index.tsx
BtencateSphereon: the result come from the passed in credentials. so what ever you put in we also should get out. i think this is fine

src/screens/CredentialOverviewShareScreen/index.tsx
BtencateSphereon: see other comment

src/components/views/CredentialSelectView/index.tsx
BtencateSphereon: changed

src/screens/CredentialOverviewShareScreen/index.tsx
BtencateSphereon: changed

src/components/views/CredentialSelectView/index.tsx
BtencateSphereon: added

src/components/views/CredentialSelectView/index.tsx
BtencateSphereon: added

sanderPostma (approved):
LGTM

